PR #1, opened 2023-11-14
command: add command to allow purge-osd

subhamkrai:
Adding command to allow osd purge. Before removing osd it validates if ceph status is 'active+clean' and also wait for ack to confirm the osd removal before proceeding. Also adding Makefile to build the binary.


pkg/purge_osd/purge_osd.go
```diff
@@ -0,0 +1,33 @@
...
+	cephcommands "github.com/red-hat-storage/odf-cli/pkg/ceph"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/rook/kubectl-rook-ceph/pkg/rook"
+)
+
+func PurgeOsd(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace, osdID, forceflagValue string) {
+	var answer string
+	logging.Warning("Are you sure you want to purge osd ID %s? If so, enter 'yes'\n", osdID)
+	fmt.Scanf("%s", &answer)
```
travisn, line 16: As much as possible let's change the behavior of the commands in the upstream tool. In this case, we can improve the purge-osd implementation upstream so that it prompts to really purge the OSD. It could also detect first if the OSD can safely be removed without the prompt, before the prompt is shown. Then we don't need to implement this prompt in the downstream tool.

pkg/purge_osd/purge_osd.go
```diff
@@ -0,0 +1,33 @@
...
+	if err := PromptToContinueOrCancel(answer, "yes"); err != nil {
+		logging.Fatal(fmt.Errorf("Purge osd osd.%s %s ", osdID, err))
+	}
+
+	if err := cephcommands.CheckPgStatus(ctx, clientsets, operatorNamespace, cephClusterNamespace); err != nil {
```
travisn, line 21: I don't think we need to check PG status. The upstream tool should just check if it's `safe-to-destroy`. And if it's not safe, then we prompt in the upstream tool if the user wants to force it.

pkg/ceph/ceph.go
```diff
@@ -0,0 +1,57 @@
+package cephcommands
+
+import (
```
travisn, line 3: As much as possible, I would hope that we can avoid implementation under `pkg/` since we expect the implementation to be in the upstream tool. The `odf` tool hopefully is just a thin `cmd/` wrapper around the upstream tool, at least for the rook and ceph commands.

cmd/commands/root.go
```diff
@@ -0,0 +1,79 @@
...
+	KubeConfig           string
+	KubeContext          string
+	OperatorNamespace    string
+	CephClusterNamespace string
+)
+
+// RootCmd represents the odf command
+var RootCmd = &cobra.Command{
+	Use:              "odf",
+	Short:            "Odf provides common management and troubleshooting tools for ODF clusters.",
```
travisn, line 24:
```suggestion
	Short:            "Management and troubleshooting tools for ODF clusters.",
```

travisn (changes requested):

pkg/purge_osd/purge_osd.go
subhamkrai: In upstream when we trigger command `ceph osd remove ...` the rook command to remove the osd we first mark osd out and the if osd is `safe-to-destroy` here https://github.com/rook/rook/blob/master/pkg/daemon/ceph/osd/remove.go#L90 so in that case we should not need this to check again.

@travisn  The reason I added the prompt and check pg status as these were requested from support team to add these pre check.

Makefile
```diff
@@ -0,0 +1,16 @@
+build:
```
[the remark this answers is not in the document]
subhamkrai: added separate commit.

cmd/commands/purge_osd.go
```diff
@@ -0,0 +1,41 @@
...
+
+var PurgeCmd = &cobra.Command{
+	Use:   "purge",
+	Short: "Call subcommands like 'osd'",
+	Args:  cobra.ExactArgs(1),
+}
+
+var osdCmd = &cobra.Command{
+	Use:   "osd",
+	Short: "Permanently remove an OSD from the cluster. For example, purge-osd 0",
```
travisn, line 21:
```suggestion
	Short: "Permanently remove an OSD from the cluster. For example, purge osd 0",
```

cmd/commands/purge_osd.go
```diff
@@ -0,0 +1,41 @@
...
+
+	purgeosd "github.com/red-hat-storage/odf-cli/pkg/purge_osd"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/spf13/cobra"
+)
+
+var PurgeCmd = &cobra.Command{
+	Use:   "purge",
+	Short: "Call subcommands like 'osd'",
```
travisn, line 15: As long as only one subcommand we could have specific help text
```suggestion
	Short: "Purge an OSD",
```

cmd/commands/root.go
```diff
@@ -0,0 +1,79 @@
+package command
```
travisn, line 1: Could we actually break this out in a separate PR? The first PR could have the makefile and just the most basic command like printing help. Then we can have follow-up PRs to start adding each command.

cmd/commands/root.go
```diff
@@ -0,0 +1,79 @@
...
+	KubeContext          string
+	OperatorNamespace    string
+	CephClusterNamespace string
+)
+
+// RootCmd represents the odf command
+var RootCmd = &cobra.Command{
+	Use:              "odf",
+	Short:            "Management and troubleshooting tools for ODF clusters.",
+	Aliases:          []string{"fdf"},
```
travisn, line 25: How does the alias work on the root command? I thought we would have to rename the binary or have a symlink to get the `fdf` cli tool name.

cmd/commands/root.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	cobra.CheckErr(RootCmd.Execute())
+}
+
+func init() {
+	RootCmd.PersistentFlags().StringVar(&KubeConfig, "kubeconfig", "", "Openshift config path")
+	RootCmd.PersistentFlags().StringVar(&OperatorNamespace, "operator-namespace", "", "Openshift namespace where ODF operator is running")
```
travisn, line 42:
```suggestion
	RootCmd.PersistentFlags().StringVar(&OperatorNamespace, "operator-namespace", "", "Openshift namespace where the ODF operator is running")
```

cmd/commands/root.go
```diff
@@ -0,0 +1,79 @@
...
+	RootCmd.PersistentFlags().StringVarP(&CephClusterNamespace, "namespace", "n", "openshift-storage", "Openshift namespace where StorageCluster is created")
```
travisn, line 43:
```suggestion
	RootCmd.PersistentFlags().StringVarP(&CephClusterNamespace, "namespace", "n", "openshift-storage", "Openshift namespace where the StorageCluster CR is created")
```

cmd/commands/root.go
```diff
@@ -0,0 +1,79 @@
...
+	RootCmd.PersistentFlags().StringVar(&KubeContext, "context", "", "Openshift context to use")
+}
+
+func GetClientsets(ctx context.Context) *k8sutil.Clientsets {
+	var err error
+
+	clientsets := &k8sutil.Clientsets{}
```
travisn, line 50: Can we just call the `k8sutil.GetClientsets()` method? Why do we need to duplicate the method here? We're anyway using that `Clientsets` struct.

pkg/purge_osd/purge_osd.go
travisn: Why not just make these improvements in the upstream tool? I don't see why it's specific to downstream.

The whole purpose of the `safe-to-destroy` is to check if the PGs would allow removing the OSD safely. If the PGs aren't clean enough to allow the OSD to be safely removed, then we wouldn't remove it without the force flag. So I don't see why the additional PG check helps. Perhaps the more important question is whether we allow the `--force` flag to proceed, or if we would require prompting to type something like `yes-really-force-osd-removal`.

travisn (changes requested):

pkg/purge_osd/purge_osd.go
subhamkrai: >So I don't see why the additional PG check helps. Perhaps the more important question is whether we allow the --force flag to proceed, or if we would require prompting to type something like yes-really-force-osd-removal.

This is valid question I guess, I'll try get answer from Bipin.

cmd/commands/root.go
subhamkrai: I agree splitting will make the PR easier to review and also for me to make review changes. Here is the #3 PR.

subhamkrai: Let's get the pr #3 ready and merge and then we'll come back to this PR and any other PR for reviews and changes.

README.md
```diff
@@ -6,6 +6,7 @@ The ODF CLI tool provides configuration and troubleshooting commands for OpenShi
 
 - `odf set recovery-profile <profile>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options `high_client_ops`, `high_recovery_ops`, or `balanced`. The default is `balanced`.
 - `odf get recovery-profile`: Get the recovery profile value.
+- `odf rook purge-osd <Id>`: Permanently remove an OSD from the cluster.
```
travisn, line 9: I thought this was the command in the design doc? 
```suggestion
- `odf purge osd <Id>`: Permanently remove an OSD from the cluster.
```

travisn (changes requested):

README.md
subhamkrai: I liked how the `purge-osd` goes with `rook`  as the parent sub-command and with time it will be easier to add more under rook I guess, similar to upstream. Thoughts?

travisn: According to the design doc, we weren't going to focus on commands under component names, to keep more consistency with the pattern: `odf <verb> <sub-verb/subject> <params>`.

cmd/odf/rook/purge_osd.go
```diff
@@ -0,0 +1,38 @@
...
+	Short:   "Permanently remove an OSD from the cluster.",
+	Args:    cobra.ExactArgs(1),
+	Example: "odf purge-osd <id>",
+	PreRunE: validateOsdID,
+	Run: func(cmd *cobra.Command, args []string) {
+		// ctx := cmd.Context()
+		// clientsets := root.GetClientsets(cmd.Context())
+		// forceflagValue := cmd.Flag(forceFlag).Value.String()
+
+		//rook.PurgeOsd(ctx, clientsets, root.OperatorNamespace, root.StorageClusterNamespace, args[0], forceflagValue)
```
travisn, line 23: The `--prompt-if-not-safe` flag needs to be passed here unless they specified the force flag.

travisn: But perhaps we don't even need to support the force flag from the downstream CLI tool, if we always want them to confirm. That's a good question for support team.

README.md
travisn: Updated the design doc with more thoughts on this topic... @subhamkrai I was discussing with @jarrpa and he pointed out that purging an OSD is really specific to ceph and really warrants the component name. So your original suggestion for `odf rook purge-osd` was good, it just took me a while to come around. I would just suggest instead the `ceph` component name since OSDs really belong to ceph, even if we do implement it in the rook operator, to make it:
```
odf ceph purge-osd <id>
```

cmd/odf/rook/purge_osd.go
subhamkrai: Last, I checked support team wanted to get ack and IMO it's safe to get ack every time.

README.md
subhamkrai: yeah having a component name with `purge-osd` make sense

subhamkrai:
Testing result
```yaml
srai@fedora ~ (add-osd-purge-cmd) $ ./bin/odf  -n rook-ceph ceph purge-osd 0 --force
Error: command terminated with exit code 16
Error EBUSY: OSD(s) 0 have 32 pgs currently mapped to them. 
Warning: Are you sure you want to purge-osd osd.0? This will possible lead to data loss. If yes enter 'yes-i-really-want-to-destroy'
yes-i-really-want-to-destroy
Info: Running purge osd command
2023-12-07 15:21:19.390641 W | cephcmd: loaded admin secret from env var ROOK_CEPH_SECRET instead of from file
2023-12-07 15:21:19.390799 I | rookcmd: starting Rook v1.13.0-alpha.0.28.g6fba73621 with arguments 'rook ceph osd remove --osd-ids=0 --force-osd-removal=true'
2023-12-07 15:21:19.390815 I | rookcmd: flag values: --force-osd-removal=true, --help=false, --log-level=INFO, --osd-ids=0, --preserve-pvc=false
2023-12-07 15:21:19.390824 I | ceph-spec: parsing mon endpoints: a=10.107.111.121:6789
2023-12-07 15:21:19.402162 I | cephclient: writing config file /var/lib/rook/rook-ceph/rook-ceph.config
2023-12-07 15:21:19.402692 I | cephclient: generated admin config in /var/lib/rook/rook-ceph
2023-12-07 15:21:19.831121 I | cephosd: validating status of osd.0
2023-12-07 15:21:19.831133 I | cephosd: osd.0 is healthy. It cannot be removed unless it is 'down'
```

cmd/odf/ceph/purge_osd.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			logging.Fatal(errors.Wrapf(err, "failed to parse --force-osd-removal flag"))
+		}
+
+		safeToDestroy, err := osd.SafeToDestroy(ctx, clientsets, root.OperatorNamespace, root.StorageClusterNamespace, args[0])
+		if !safeToDestroy && !forceOSDRemovalBool {
+			logging.Fatal(errors.Wrapf(err, "osd.%s is not safe to destroy", args[0]))
+		} else if !safeToDestroy && forceOSDRemovalBool {
+			var answer string
+			logging.Warning("Are you sure you want to purge-osd osd.%s? This will possible lead to data loss. If yes enter 'yes-i-really-want-to-destroy'", args[0])
```
travisn, line 39:
```suggestion
			logging.Warning("Are you sure you want to purge osd.%s? The OSD is *not* safe to destroy. This may lead to data loss. If you are sure the OSD should be purged, enter 'yes-force-destroy-osd'", args[0])
```

travisn (changes requested):
The output of that test looks good, when we detect if it's not safe to destroy and the ask for the prompt in that case. One modification suggested... There is no need for the `--force` flag in the `odf` cli. We will just always prompt if it's not safe to destroy, and there is no need to pass the force flag. It doesn't really force the removal. The prompt is what forces the removal.

subhamkrai:
Testing result when cancelled:

```yaml
 ./bin/odf -n rook-ceph ceph purge-osd 0
Error: command terminated with exit code 11
Error EAGAIN: 1 pgs have unknown state; cannot draw any conclusions
Warning: Are you sure you want to purge osd.0? The OSD is *not* safe to destroy. This may lead to data loss. If you are sure the OSD should be purged, enter 'yes-force-destroy-osd'
no
Error: cancelled
```

Test result when continued:

```yaml
./bin/odf -n rook-ceph ceph purge-osd 0
Error: command terminated with exit code 11
Error EAGAIN: 1 pgs have unknown state; cannot draw any conclusions
Warning: Are you sure you want to purge osd.0? The OSD is *not* safe to destroy. This may lead to data loss. If you are sure the OSD should be purged, enter 'yes-force-destroy-osd'
yes-force-destroy-osd
Info: Running purge osd command
2023-12-08 03:22:31.172268 W | cephcmd: loaded admin secret from env var ROOK_CEPH_SECRET instead of from file
2023-12-08 03:22:31.172675 I | rookcmd: starting Rook v1.13.0-alpha.0.28.g6fba73621 with arguments 'rook ceph osd remove --osd-ids=0 --force-osd-removal=true'
2023-12-08 03:22:31.172803 I | rookcmd: flag values: --force-osd-removal=true, --help=false, --log-level=INFO, --osd-ids=0, --preserve-pvc=false
2023-12-08 03:22:31.172921 I | ceph-spec: parsing mon endpoints: a=10.107.111.121:6789
2023-12-08 03:22:31.179494 I | cephclient: writing config file /var/lib/rook/rook-ceph/rook-ceph.config
2023-12-08 03:22:31.179599 I | cephclient: generated admin config in /var/lib/rook/rook-ceph
2023-12-08 03:22:31.530304 I | cephosd: validating status of osd.0
2023-12-08 03:22:31.530372 C | rookcmd: failed to get osd status for osd 0: not found osd.0 in OSDDump
Error: command terminated with exit code 1
```

cmd/odf/ceph/ceph.go
```diff
@@ -0,0 +1,16 @@
+package rook
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// CephCmd represents the set command
+var CephCmd = &cobra.Command{
+	Use:                "ceph",
+	Short:              "Calls subcommands like 'purge-osd'",
```
travisn, line 10:
```suggestion
	Short:              "Commands for configuring Ceph",
```

travisn (changes requested):

docs/purge-osd.md
```diff
@@ -0,0 +1,41 @@
+# Ceph
```
travisn, line 1: This file needs to be renamed to ceph.md

travisn (changes requested):
just one more nit

travisn (changes requested):

travisn (approved):

PR #2, opened 2023-11-14
command: command to allow changing mclock profile

subhamkrai:
command: command to allow changing mclock profile
Adding command to choose/modify from ODF the mclock scheduler
profile that is being used during recovery. We'll allow to update values
from 'high_client_ops, high_recovery_ops, balanced'. We also allow
passing option argument name of the specific osd where the settings will
be allowed and by default it will allow to all the osds.

Signed-off-by: subhamkrai <srai@redhat.com>


subhamkrai:
This pr is on top of #1

cmd/commands/root.go
```diff
@@ -38,3 +47,50 @@ func init() {
...
+}
+
+func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, storageClusterNamespace string) {
+	_, err := k8sclientset.Kube.CoreV1().Namespaces().Get(ctx, operatorNamespace, v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Operator namespace '%s' does not exist. %v", operatorNamespace, err))
+	}
+	_, err = k8sclientset.Kube.CoreV1().Namespaces().Get(ctx, storageClusterNamespace, v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("CephCluster namespace '%s' does not exist. %v", storageClusterNamespace, err))
```
travisn, line 94:
```suggestion
		logging.Fatal(fmt.Errorf("StorageCluster namespace '%s' does not exist. %v", storageClusterNamespace, err))
```

cmd/commands/set.go
```diff
@@ -0,0 +1,61 @@
...
+	Short:              "Call subcommands like 'recover-profile'",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+}
+
+var validArgs = []string{"high_client_ops", "high_recovery_ops", "balanced"}
+
+var recoverProfile = &cobra.Command{
+	Use:                "recover-profile",
+	Short:              "Update the mclock scheduler profile value `high_client_ops, high_recovery_ops, balanced` used during recovery for single osd or all the osds",
```
travisn, line 24: Do we really need the flexibility to set the recovery profile for a single OSD? I thought the downstream cli would just need to set a single global setting so all OSDs are affected. Unless support requested that feature, I would vote to remove that option.
```suggestion
	Short:              "Set the recovery profile to favor new IO, recovery, or balanced mode with options high_client_ops, high_recovery_ops, or balanced",
```

cmd/commands/set.go
```diff
@@ -0,0 +1,61 @@
...
+
+// SetCmd represents the set command
+var SetCmd = &cobra.Command{
+	Use:                "set",
+	Short:              "Call subcommands like 'recover-profile'",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+}
+
+var validArgs = []string{"high_client_ops", "high_recovery_ops", "balanced"}
```
travisn, line 20:
```suggestion
var validRecoveryOptions = []string{"high_client_ops", "high_recovery_ops", "balanced"}
```

cmd/commands/set.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	},
+}
+
+func init() {
+	SetCmd.AddCommand(recoverProfile)
+}
+
+func isValidArg(args []string) bool {
+	for c, valid := range validArgs {
```
travisn, line 42: Assigning a var to args[0] will really help with the readability of this method
```suggestion
	recoveryInput := args[0]
	for c, valid := range validArgs {
```

cmd/commands/set.go
```diff
@@ -0,0 +1,61 @@
...
+		if args[0] == valid {
+			break
+		} else if c == len(validArgs)-1 {
+			logging.Warning("Invalid argument %s, valid arguments are: %v\n", args[0], validArgs)
+			return false
+		}
+	}
+
+	if len(args) == 2 {
```
travisn, line 51: Hopefully no need for this scenario

travisn (changes requested):

travisn:
With each command, let's also add it to the readme, similar to the upstream tool docs. This would provide the basis for the downstream docs.

cmd/commands/set.go
subhamkrai: I don't have opinion here, I found the option in the ceph docs so I added it but I'm okay to remove the option to pass osd.

subhamkrai:
> With each command, let's also add it to the readme, similar to the upstream tool docs. This would provide the basis for the downstream docs.

added

sp98:
recommending a change in the directory structure.
``` cmd
        set
          cmd.go
           mclock.go
        get 
           cmd.go
           health.go
           drHealth.go
        purge
           cmd.go
           osd.go
   ```

**Review by sp98 — CHANGES_REQUESTED** [review]
some nits

File: cmd/commands/set/set.go [hunk 6]
```diff
@@ -0,0 +1,17 @@
+package set
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// SetCmd represents the set command
+var SetCmd = &cobra.Command{
+	Use:                "set",
+	Short:              "Call subcommands like 'recovery-profile'",
```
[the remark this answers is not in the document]
subhamkrai: `Short` is printed in the terminal when the user runs `odf set`  so, we followed this approach to mention the `sub-verb` in the `verb` and `sub-verb` will explain what command does.

README.md
```diff
@@ -1,19 +1,66 @@
...
+4. `--context`: the name of the Openshift context to be used (optional).
+
+    ```bash
+    odf --context=$(oc config current-context) [commands]
+    ```
 
 ## Commands
 
-TBD
+- `set <args>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options high_client_ops, high_recovery_ops, or balanced
```
BlaineEXE, line 39: Is this intended to allow setting any ceph config value, or just the mclock profile? Is this more accurate to the implementation?

```suggestion
- `set recovery-profile <profile>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options high_client_ops, high_recovery_ops, or balanced
```

pkg/set/set.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+)
+
+func SetProfile(ctx context.Context, clientsets *k8sutil.Clientsets, recoveryOption, operatorNamespace, storageClusterNamespace string) {
+
+	cephArgs := []string{"config", "set", "osd", "osd_mclock_profile", recoveryOption}
+
+	exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", cephArgs, operatorNamespace, storageClusterNamespace, true, false)
```
BlaineEXE, line 14: I'm curious why we are taking this approach rather than allowing users to use `ceph config set osd osd_mclock_profile <option>` directly. 

It seems like this feature is a one-to-one passthrough where the CLI is just renaming one config to another. The CLI is effectively acting as an allow-list for configurations that users can set.

Will we be playing catch-up to implement `odf-cli set <option>` for any Ceph tunable that users want to set if we only implement some of them?

BlaineEXE, line 14: We have a recent Rook design doc that intends allow users to specify Ceph config changes on the CephCluster itself. If that feature were available in Rook, would the CLI:
1. deprecate this command and tell users to use the CephCluster config spec?
2. modify this command to use the CephCluster config spec?
3. do neither?

Do we need to consider the answer to this question before proceeding here?

See: https://github.com/rook/rook/pull/13202

docs/set.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Recovery-profile
+
+This command is used choose/modify the [mclock](https://docs.ceph.com/en/quincy/rados/configuration/mclock-config-ref/#mclock-profiles-definition-and-purpose) scheduler profile that is being used during recovery.
+
+We can use the following built-in profile types:
+
+- [balanced](https://docs.ceph.com/en/quincy/rados/configuration/mclock-config-ref/#balanced-default) (default)
+- [high_client_ops](https://docs.ceph.com/en/quincy/rados/configuration/mclock-config-ref/#high-client-ops)
+- [high_recovery_ops](https://docs.ceph.com/en/quincy/rados/configuration/mclock-config-ref/#high-recovery-ops)
```
BlaineEXE, line 15: Should these link to RHCS docs?

README.md
BlaineEXE, line 39: Do we intend to implement an `unset` command to revert options back to their defaults?

pkg/set/set.go
subhamkrai: I see, but currently, that is in cephCluster cr not in storageCluster cr which is what ocs operator will reconcile. Also, the main reason for only individual ceph commands is because we discussed that we don't want to expose all the ceph options to customers only add what they request for or want.

Also, if in the future we add in storageCluster cr to allow modifying all the ceph config then in this command we can update the command to patch storageCluster cr but this will require discussion with the support team also what the customer wants as currently customer requests was a tool like this as they were not interested in running the commands in the toolbox.

subhamkrai: Also, somehow I feel the rook feature to modify the ceph config should be in configmap similar to `rook-ceph-operator-config` configmap where we add operator and CSI settings rather than in cephCluster CR.  IMO we'll make cephCluster CR too big and complicated.

README.md
subhamkrai: this is intended to allow only mclock profile config.

pkg/set/set.go
subhamkrai: >I'm curious why we are taking this approach rather than allowing users to use ceph config set osd osd_mclock_profile <option> directly.

The request from the customer was they didn't want to run the command from the toolbox they wanted something from odf.

>Will we be playing catch-up to implement odf-cli set <option> for any Ceph tunable that users want to set if we only implement some of them?

I think the purpose of adding this command is we decided to not expose all the ceph config options to customers directly and restrict customers to what config they can set based on their requests.

Also, what I have heard is that the toolbox is not always available to customers.

README.md
subhamkrai: Maybe in the future not planned in the near roadmap

docs/set.md
subhamkrai: I was not able to point url to each individual topics

pkg/set/set.go
BlaineEXE: Thanks for answering this and my other questions. Those reasons all make sense.

cmd/odfcli/set/recovery-profile_test.go
```diff
@@ -0,0 +1,25 @@
...
+		args string
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{"Valid Input with 'high_client_ops' as arg", args{args: "high_client_ops"}, true},
+		{"Valid Input with 'balanced' as arg", args{args: "balanced"}, true},
+		{"Invalid input for mclock profile", args{args: "balance"}, false},
```
BlaineEXE, line 16: Should probably test that all valid inputs are accepted.
Also, empty string test for corner case.

BlaineEXE (commented):
Unit testing can only validate this so deeply. Are there plans to do an e2e integration test to ensure this sets the value as expected on a running cluster?

travisn (changes requested):

cmd/odfcli/root.go
```diff
@@ -0,0 +1,96 @@
+package odfcli
```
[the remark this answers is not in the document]
subhamkrai: I got the suggestion from @sp98  and this directory helps in collaborating and expanding commands and easier to collaborate when multiple are working on the same sub-command. Also, this is a common directory structure seen in other downstream cli tools.

@sp98 would you like to add anything?

README.md
```diff
@@ -1,19 +1,66 @@
...
+### Command to set mclock recovery profile
+
+```code
+odf -n openshift-storage set recovery-profile high_client_ops
+```
+
+To verify the mclock recovery profile settings
+
+> ```text
+> ceph config get osd
```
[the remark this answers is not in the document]
subhamkrai: added

cmd/odfcli/set/recovery-profile.go
```diff
@@ -0,0 +1,41 @@
...
+
+var validRecoveryOptions = []string{"high_client_ops", "high_recovery_ops", "balanced"}
+
+var recoveryProfile = &cobra.Command{
+	Use:                "recovery-profile",
+	Short:              "Set the recovery profile to favor new IO, recovery, or balanced mode with options high_client_ops, high_recovery_ops, or balanced",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		recoveryInput := args[0]
```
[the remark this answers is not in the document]
subhamkrai: added

subhamkrai:
This is how it looks now
```bash
srai@192 ~ (add-cmd-mclock) $ ./bin/odf 
Management and troubleshooting tools for ODF clusters.

Usage:
  odf [command]

Available Commands:
  get         Get ODF configuration
  help        Help about any command
  set         Set ODF configuration

Flags:
      --context string              Openshift context to use
  -h, --help                        help for odf
      --kubeconfig string           Openshift config path
  -n, --namespace string            Openshift namespace where the StorageCluster CR is created (default "openshift-storage")
      --operator-namespace string   Openshift namespace where the ODF operator is running

Use "odf [command] --help" for more information about a command.
~/go/src/github.com/odf-cli
srai@192 ~ (add-cmd-mclock) $ ./bin/odf -n rook-ceph get recovery-profile
Info: high_recovery_ops
~/go/src/github.com/odf-cli
srai@192 ~ (add-cmd-mclock) $ ./bin/odf -n rook-ceph set recovery-profile high_client_ops
~/go/src/github.com/odf-cli
srai@192 ~ (add-cmd-mclock) $ ./bin/odf -n rook-ceph get recovery-profile
Info: high_client_ops
~/go/src/github.com/odf-cli
srai@192 ~ (add-cmd-mclock) $ ./bin/odf -n rook-ceph set recovery-profile
~/go/src/github.com/odf-cli
srai@192 ~ (add-cmd-mclock) $ ./bin/odf -n rook-ceph get recovery-profile
Info: balanced
~/go/src/github.com/odf-cli
srai@192 ~ (add-cmd-mclock) $ 
```

subhamkrai:
@travisn @BlaineEXE @sp98 since the pr is getting bigger, I would happy to split the changes into 3 commits
1. commit for the command  `set`
2.  commit for the command  `get`
3. commit for the docs

let me know

README.md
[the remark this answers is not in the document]
BlaineEXE: This is a really good point. Users may forget what they have done and want to see the current strategy, so `get` will be very important for them. Moreso than `unset` I suspect.

cmd/odfcli/set/recovery-profile.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/red-hat-storage/odf-cli/cmd/odfcli"
+	"github.com/red-hat-storage/odf-cli/pkg/set"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/spf13/cobra"
+)
+
+var validRecoveryOptions = []string{"high_client_ops", "high_recovery_ops", "balanced"}
+
+var recoveryProfile = &cobra.Command{
```
BlaineEXE, line 14: For complex command structures that develop over time, it's good to include the parent in the subcommand name. You can look at Kubectl's command structure, for reference: https://github.com/kubernetes/kubectl/tree/master/pkg/cmd/create

```suggestion
var setRecoveryProfile = &cobra.Command{
```

cmd/odfcli/set/recovery-profile.go
```diff
@@ -0,0 +1,47 @@
...
+	Use:                "recovery-profile",
+	Short:              "Set the recovery profile to favor new IO, recovery, or balanced mode with options high_client_ops, high_recovery_ops, or balanced",
+	DisableFlagParsing: true,
+	Args:               cobra.MaximumNArgs(1),
```
BlaineEXE, line 18: I have 2 thoughts here:
1. My opinion is that this should be cobra.ExactArgs(1). I don't think the behavior of setting the default when the user gives no setting value matches with how most CLIs work. Users will generally assume that `cli set key` with no value is a typo.
2. Cobra offers more features for setting string arg values that would cut down on the boilerplate and offer other benefits:

Cobra has a `ValidArgs` property that we should use here. This allows us to totally skip the `isValidArg()` boilerplate code, corresponding unit tests, as well as usage of the custom function in the `Command.Run()` method. Pair it with the `MatchAll(ExactArgs(n), OnlyValidArgs)` for validation.

e.g.
```
myCmd := &cobra.Command{
    # ...
    ValidArgs: []string{"balanced", "etc"},
    Args:      cobra.MatchAll(cobra.ExactArgs(1), cobra.OnlyValidArgs)
}
```

FWIW, this would also allow cobra to create auto-completion for the command more easily if we decide to enable it in the future.

cmd/odfcli/set/recovery-profile.go
```diff
@@ -0,0 +1,47 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		var recoveryInput string
+		if len(args) == 0 {
+			recoveryInput = "balanced"
+		} else {
+			recoveryInput = args[0]
+		}
```
BlaineEXE, line 25: Related to my other comment. I think the behavior of re-setting the default when no value is given is one that could cause a problem for a user who intends to set a value but forgets to include it. 

Consider this scenario:
- I am a support engineer working with a customer who needs to speed up Ceph recovery ops
- I advise the customer to use `odf-cli set recover-profile` to set high recovery ops
- The customer misinterprets this, and only runs `odf-cli set recover-profile`
- The command executes successfully, leading the customer to believe they have done the right thing
- The customer tells the support engineer they did it
- The customer's issue still persists, and the support engineer, believing the customer did the right thing, proceeds to debug other things that are not the root cause

If, instead, we make it so the value selection is mandatory, the customer cannot accidentally apply the default unexpectedly, and we could save customers, support engineers, or ourselves a lot of time in the future.

@travisn thoughts?

BlaineEXE:
> since the pr is getting bigger, I would happy to split the changes into 3 commits
> 
> 1. commit for the command  `set`
> 2. commit for the command  `get`
> 3. commit for the docs
> 
> let me know

I am of the opinion that every commit that has a feature should also include tests and docs as part of the commit. In practice, this allows features to be reverted more easily. If the commands are added individually, and the doc commit has docs for both changes, it becomes harder to revert only a single one of the commits.

I understand that this isn't really a practical concern for us right now, but I believe that it is good to always practice good commit hygiene so that we don't forget in the future when it is important.

By my logic, I think it is appropriate for `set` and `get` to be 2 separate commits, but each should introduce their own documentation. 

I often find that it is challenging to keep separate commits that all involve docs and tests clean when working locally. I prefer to make separate PRs when that is the case to avoid the headache. Personally, I think `get` is sufficiently small and sufficiently related to the feature that it makes sense to do both at the same time, but that's your call.

In addition, a lot of the "bigness" is the go mod and sum files, which I don't count as part of the code review, and I have given some other suggestions for how to reduce the amount of code being written here by taking advantage of more of Cobra's features.

BlaineEXE:
You showed an output snippet here, and I notice that `help` breaks up the `get` and `set` commands. When I look at the output of `kubectl`, I don't actually see `help` listed as a subcommand, even though it is allowed. I think it's safe to hide the help command from the usage text if we want to. 

```console
srai@192 ~ (add-cmd-mclock) $ ./bin/odf 
Management and troubleshooting tools for ODF clusters.

Usage:
  odf [command]

Available Commands:
  get         Get ODF configuration
  help        Help about any command
  set         Set ODF configuration
# ...
```

[This thread](https://github.com/spf13/cobra/issues/587#issuecomment-810159087) suggests using `rootCmd.SetHelpCommand(&cobra.Command{Hidden: true})` to quickly hide the builtin help command, if you'd like to try it.

Alternately, we could use command groups to group get and set together, but that seems like more work than is helpful right now.

cmd/odfcli/root.go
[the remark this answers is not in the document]
travisn: If we follow the structure similar to kubectl as @BlaineEXE pointed out, I'd suggest `root.go` in `cmd/`, and we have subdirectories such as `set`, `get`, etc.

cmd/odfcli/set/recovery-profile.go
travisn: Sounds good to keep it required, agreed it's better this way to avoid unintended settings.

pkg/set/set.go
```diff
@@ -0,0 +1,15 @@
+package set
```
travisn, line 1: For directory structure under `cmd/` it makes sense to have `set/` or `get/` subfolders. But under `pkg/` I think we can have a different directory structure that is more convenient for each component. For example if we had `pkg/rook/osd/recovery.go`, then related methods for recovery Get and Set could be included in the same implementation go file, which makes it simpler to share arguments or helpers for the same feature.

travisn (changes requested):

pkg/set/set.go
BlaineEXE: +1

cmd/odfcli/set/set.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+// SetCmd represents the set command
+var SetCmd = &cobra.Command{
+	Use:                "set",
+	Short:              "Set ODF configuration",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
```
BlaineEXE, line 12: I think the `Args` field doesn't matter here. From what I understand of Cobra, subcommands don't count as arguments. For the multus CLI in Rook, which is comparable, I left this field blank

https://github.com/rook/rook/blob/0f6e7ee921e98776c31b4b0912e69f093caeccfc/cmd/rook/userfacing/multus/multus.go#L30-L41

subhamkrai:
> Unit testing can only validate this so deeply. Are there plans to do an e2e integration test to ensure this sets the value as expected on a running cluster?

Yes, we'll add the CI similar to upstream version for our e2e test

cmd/odfcli/root.go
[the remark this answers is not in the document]
subhamkrai: we can't keep `main.go` and `root.go` in under same directory, it will import cycle issue.  What we can do is keep `main.go` in root(similar to go.mod) directory and keep `root.go` under `cmd/` I have seen this directory structure also.

cmd/odfcli/set/recovery-profile.go
subhamkrai: I agree and I explored the `ValidArgs` but I was not aware of `MatchAll`.

cmd/odfcli/set/recovery-profile.go
subhamkrai: done.

subhamkrai:
> > since the pr is getting bigger, I would happy to split the changes into 3 commits
> > 
> > 1. commit for the command  `set`
> > 2. commit for the command  `get`
> > 3. commit for the docs
> > 
> > let me know
> 
> I am of the opinion that every commit that has a feature should also include tests and docs as part of the commit. In practice, this allows features to be reverted more easily. If the commands are added individually, and the doc commit has docs for both changes, it becomes harder to revert only a single one of the commits.
> 
> I understand that this isn't really a practical concern for us right now, but I believe that it is good to always practice good commit hygiene so that we don't forget in the future when it is important.
> 
> By my logic, I think it is appropriate for `set` and `get` to be 2 separate commits, but each should introduce their own documentation.
> 
> I often find that it is challenging to keep separate commits that all involve docs and tests clean when working locally. I prefer to make separate PRs when that is the case to avoid the headache. Personally, I think `get` is sufficiently small and sufficiently related to the feature that it makes sense to do both at the same time, but that's your call.
> 
> In addition, a lot of the "bigness" is the go mod and sum files, which I don't count as part of the code review, and I have given some other suggestions for how to reduce the amount of code being written here by taking advantage of more of Cobra's features.


In that case going with single commit only, it will be also easier for me

cmd/odfcli/set/set.go
subhamkrai: right, cobra don't count subcommands as args

pkg/set/set.go
subhamkrai: done

docs/get.md
```diff
@@ -0,0 +1,19 @@
+# Get
+
+The get command supports the following sub-commands:
+
+* [Recovery-profile](#recovery-profile)
```
travisn, line 5:
```suggestion
* [recovery-profile](#recovery-profile)
```

docs/get.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Recovery-profile
```
travisn, line 7:
```suggestion
## recovery-profile
```

docs/get.md
```diff
@@ -0,0 +1,19 @@
...
+
+This commands will display the recovery-profile value set for the osd.
```
travisn, line 9:
```suggestion
This command will display the recovery-profile value set for the osd.
```

docs/get.md
```diff
@@ -0,0 +1,19 @@
...
+
+### Example
```
travisn, line 11: Let's not have a subheader, just show the example directly.
```suggestion
```

docs/get.md
```diff
@@ -0,0 +1,19 @@
...
+
+```bash
+odf get recovery-profile
+```
+
+> ```text
+> Info: high_recovery_ops
```
travisn, line 18: `get` commands should directly print the raw results, without the "Info:" prefix.

docs/set.md
```diff
@@ -0,0 +1,22 @@
+# Set
+
+The set command supports the following sub-commands:
+
+* [Recovery-profile](#recovery-profile)
```
travisn, line 5:
```suggestion
* [recovery-profile](#recovery-profile)
```

docs/set.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Recovery-profile
```
travisn, line 7:
```suggestion
## recovery-profile
```

docs/set.md
```diff
@@ -0,0 +1,22 @@
...
+
+This command will set the recovery profile to favor either the new client IO, recovery IO, or a balanced mode. The default is `balanced` mode.
+We can use the following built-in profile types:
+
+* balanced(default)
+* high_client_ops
+* high_recovery_ops
+
+### Example
```
travisn, line 16:
```suggestion
```

docs/set.md
```diff
@@ -0,0 +1,22 @@
...
+The set command supports the following sub-commands:
+
+* [Recovery-profile](#recovery-profile)
+
+## Recovery-profile
+
+This command will set the recovery profile to favor either the new client IO, recovery IO, or a balanced mode. The default is `balanced` mode.
+We can use the following built-in profile types:
+
+* balanced(default)
```
travisn, line 12: Let's not mention default here since it's already mentioned in the paragraph.
```suggestion
* balanced
```

docs/set.md
```diff
@@ -0,0 +1,22 @@
...
+* high_client_ops
+* high_recovery_ops
+
+### Example
+
+```code
+odf set recovery-profile high_client_ops
+```
+
+To verify the mclock recovery profile setting run [this command](get.md#recovery-profile).
```
travisn, line 22: Let's not mention mclock, it's a detail the ODF users won't care about.
```suggestion
To verify the recovery profile setting run [this command](get.md#recovery-profile).
```

go.mod
```diff
@@ -3,17 +3,94 @@ module github.com/red-hat-storage/odf-cli
 go 1.21.0
```
travisn, line 3: required for compilation
```suggestion
go 1.21
```

cmd/odfcli/root.go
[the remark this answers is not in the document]
travisn: How about this?
```
cmd/main.go
cmd/root/root.go
cmd/get/...
cmd/set/...

**Review by travisn — CHANGES_REQUESTED** [review]
Mostly some structure, and minor doc comments

File: cmd/odfcli/root.go (hunk 11 above)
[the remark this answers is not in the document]
Comment (sp98) [reply cmd/odfcli/root.go:1]: A common pattern that is used follows `cmd/appname`  where `appname` is the name of the executable. 

https://github.com/golang-standards/project-layout/tree/master/cmd

File: docs/get.md (hunk 19 above)
Comment (subhamkrai) [reply docs/get.md:5]: this auto correct by vs code, probably due to `markdown` extension

File: docs/get.md (hunk 20 above)
Comment (subhamkrai) [reply docs/get.md:7]: this auto correct by vs code, probably due to `markdown` extension

File: docs/set.md (hunk 24 above)
Comment (subhamkrai) [reply docs/set.md:5]: this auto correct by vs code, probably due to `markdown` extension

Comment (subhamkrai) [reply docs/set.md]: never-mind, I found the fix

File: docs/get.md (hunk 20 above)
Comment (subhamkrai) [reply docs/get.md]: never-mind, I found the fix

File: docs/get.md (hunk 19 above)
Comment (subhamkrai) [reply docs/get.md]: never-mind, I found the fix

File: cmd/odfcli/root.go (hunk 11 above)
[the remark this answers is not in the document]
Comment (subhamkrai) [reply cmd/odfcli/root.go]: Going with this structure 
```
├── cmd
│   ├── main.go
│   ├── odfcli
│   │   ├── get
│   │   │   ├── get.go
│   │   │   └── get_recovery_profile.go
│   │   └── set
│   │       ├── set.go
│   │       └── set_recovery_profile.go
│   └── root
│       └── root.go
```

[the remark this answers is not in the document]
Comment (travisn) [reply cmd/odfcli/root.go]: @sp98 Nice link for the go standards, makes sense to have subdirectories to easily support multiple binaries. Not sure if we'll ever have another binary in this repo, but doesn't hurt to allow for it with this pattern. With that pattern then, seems like we need this structure so that all the go files for the binary are in its folder, and the folder is named `odf` like the tool:

```
├── cmd
│   ├── odf
│   |   ├── main.go
│   │   ├── get
│   │   │   ├── get.go
│   │   │   └── get_recovery_profile.go
│   │   └── set
│   │       ├── set.go
│   │       └── set_recovery_profile.go
│   |   └── root
│   |      └── root.go
```

File: README.md [hunk 30]
```diff
@@ -1,19 +1,44 @@
...
-
-TBD
+The ODF CLI tool provides configuration and troubleshooting commands for OpenShift Data Foundation.
 
 ## Commands
 
-TBD
+- `odf set recovery-profile <profile>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options `high_client_ops`, `high_recovery_ops`, or `balanced`. The default is `balanced`.
+- `odf get recovery-profile`: Get the recovery profile value.
+- `help` : Display help text
```
travisn, line 9:
```suggestion
- `odf help` : Display help text
```

docs/get.md
```diff
@@ -0,0 +1,19 @@
...
+
+The get command supports the following sub-commands:
+
+* [recovery-profile](#recovery-profile)
+
+## recovery-profile
+
+This command will display the recovery-profile value set for the osd.
+
+Example
```
travisn, line 11: No need for this "Example" line
```suggestion
```

docs/get.md
```diff
@@ -0,0 +1,19 @@
...
+
+```bash
+odf get recovery-profile
```
travisn, line 14: What about combining the command with the output in a single block? But if you prefer how it was, you can ignore this.
```suggestion
$ odf get recovery-profile
high_recovery_ops
```

docs/set.md
```diff
@@ -0,0 +1,22 @@
...
+* high_client_ops
+* high_recovery_ops
+
+Example
+
+```code
+odf set recovery-profile high_client_ops
+```
+
+To verify the recovery profile setting run [this command](get.md#recovery-profile).
```
travisn, line 22:
```suggestion
To verify the recovery profile setting run [odf get recovery-profile](get.md#recovery-profile).
```

travisn (changes requested):
Just a few more minor formatting suggestions, thanks!

docs/get.md
subhamkrai: I think this make sense

travisn (approved):

travisn (changes requested):
Actually, the directory structure is still not expected. See the previous discussion on that

BlaineEXE (approved):

cmd/odfcli/root.go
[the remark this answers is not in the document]
subhamkrai: done

subhamkrai:
> Actually, the directory structure is still not expected. See the previous discussion on that

done

subhamkrai:
> @BlaineEXE: changing LGTM is restricted to collaborators
> 
> In response to [this](https://github.com/red-hat-storage/odf-cli/pull/2#pullrequestreview-1757744309):
> 
> > 
> 
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.

@travisn any idea why it is showing this with Blaine LGTM, I see Blaine name owners file

sp98 (approved):

travisn (approved):

PR #3, opened 2023-11-16
core: This PR lay the foundation of the tool

subhamkrai:
adding the initial cobra go code to include
main.go and root.go based on which other cmd
will be added in future PRs. Also, adding the
Makefile.


cmd/commands/root.go
```diff
@@ -0,0 +1,40 @@
...
+	KubeConfig           string
+	KubeContext          string
+	OperatorNamespace    string
+	CephClusterNamespace string
+)
+
+// RootCmd represents the odf command
+var RootCmd = &cobra.Command{
+	Use:              "odf",
+	Short:            "Odf provides common management and troubleshooting tools for ODF clusters.",
```
travisn, line 17:
```suggestion
	Short:            "Management and troubleshooting tools for ODF clusters.",
```

cmd/commands/root.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	KubeConfig           string
+	KubeContext          string
+	OperatorNamespace    string
+	CephClusterNamespace string
```
travisn, line 11: How about the storage cluster namespace? This will be the same as the ceph namespace.
```suggestion
	StorageClusterNamespace string
```

travisn (changes requested):

cmd/commands/root.go
subhamkrai: right

travisn (approved):

travisn (commented):
/lgtm
/approve

PR #4, opened 2023-11-27
cmd: set log level of different ceph daemons

sp98:
This PR sets the log level of different ceph daemons like osd, mon and mds.


pkg/set/set.go
```diff
@@ -13,3 +13,10 @@ func SetProfile(ctx context.Context, clientsets *k8sutil.Clientsets, recoveryOpt
 
 	exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", cephArgs, operatorNamespace, storageClusterNamespace, true, false)
 }
+
+func SetLogLevel(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, storageClusterNamespace string, args []string) {
+
+	cephArgs := []string{"config", "set", args[0], args[1], args[2]}
```
travisn, line 19: We'll need some validation on these arguments so we can ensure it's only being used by the user to set log levels. Otherwise, we have just exposed a general `ceph config set` for any setting, under the alias of `log-level`. :)

travisn (changes requested):

PR #6, opened 2023-12-01
example: add example section in recovery-profile

subhamkrai:
It's nice have example mention as this will give
users idea about syntax and uses.


cmd/odf/set/set_recovery_profile.go
```diff
@@ -9,6 +9,7 @@ import (
 var setRecoveryProfile = &cobra.Command{
 	Use:                "recovery-profile",
 	Short:              "Set the recovery profile to favor new IO, recovery, or balanced mode with options high_client_ops, high_recovery_ops, or balanced",
+	Example:            "odf get recovery-profile <option>",
```
travisn, line 12:
```suggestion
	Example:            "odf set recovery-profile <option>",
```

travisn (changes requested):
For commit prefix, how about something more specific than "command", such as "example"? But let's look for patterns for commit messages used in other downstream repos and perhaps not even require a prefix.

travisn (approved):

PR #5, opened 2023-12-01
add command to set log level

sp98:
Set log level for ceph daemons

PR #7, opened 2023-12-03
cmd: validate ceph subsystems and support both log-level and memory level logs

sp98:
This is a follow up PR for #5 
- supporting both log level and memory level changes in the command. For exampel:  odf set log-level osd 5/20
- Adding support for changing logs for more subsystems


sp98:
not all of these subsystems seem to be working with the `ceph config set ..`  command.
```
sh-4.4$ ceph config set filestore debug_filestore 1/5
Error EINVAL: unrecognized config target 'filestore' ```

Trying to find out why

File: cmd/odf/set/log_level.go [hunk 1]
```diff
@@ -37,7 +37,7 @@ func validateLogLevelArgs(_ *cobra.Command, args []string) error {
 	}
 
 	// validate log level
-	match, err := regexp.MatchString("^[1-9][0-9]?$", args[1])
+	match, err := regexp.MatchString("^[1-9][0-9]?(\\/[1-9][0-9]?)?$", args[1])
```
travisn, line 40: Please add a comment that shows examples of valid values.

travisn (changes requested):
1. How about update the PR and commit title to describe the changes? You can say in the description that it's a follow-up, but it's helpful in the commit title to know what's changing, such as to point out that the log level options are being updated.
2. To follow up from [this comment](https://github.com/red-hat-storage/odf-cli/pull/5/files#r1414025462), I believe we need another argument, not just component and log level.

cmd/odf/set/log_level.go
sp98: added examples.

sp98:
> 1. How about update the PR and commit title to describe the changes? You can say in the description that it's a follow-up, but it's helpful in the commit title to know what's changing, such as to point out that the log level options are being updated.

Agree. The PR title was badly written. 

> 2. To follow up from [this comment](https://github.com/red-hat-storage/odf-cli/pull/5/files#r1414025462), I believe we need another argument, not just component and log level.
Just added list of subsystems that can be used while running a command. The command format looks like:

odf log-level <ceph component> <ceph subsystem> {log-level/memory-level} 

<ceph component> : osd, mon, mds . No validations added for this as of now. Users can pass `osd.0`, `osd.*`, etc. 
<ceph subsystem> : list of all the subsystems as shown [here](https://docs.ceph.com/en/latest/rados/troubleshooting/log-and-debug/#ceph-subsystems) 
{log-level/memory-level}: possible values  5, 5/10, 10/20 etc.

pkg/rook/common/logs.go
```diff
@@ -8,11 +8,20 @@ import (
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 )
 
-var ValidCephSubSystems = []string{"osd", "mds", "mon", "mgr", "auth"}
+var (
+	ValidCephSubSystems = []string{"default", "lockdep", "context", "crush", "mds", "mds_balancer", "mds_log", "mds_log_expire",
```
travisn, line 12: Let's not validate the subsystems. We don't want to maintain this list which will change over time.

travisn: What if an invalid subsystem is passed to ceph? Will the ceph command fail or ignore the bad setting?

pkg/rook/common/logs.go
```diff
@@ -8,11 +8,20 @@ import (
...
+		"mds_migrator", "buffer", "timer", "filer", "striper", "objecter", "rados", "rbd", "rbd_mirror", "rbd_replay", "rbd_pwl", "journaler",
+		"objectcacher", "immutable_obj_cache", "client", "osd", "optracker", "objclass", "filestore", "journal", "ms", "mon", "monc", "paxos",
+		"tp", "auth", "crypto", "finisher", "reserver", "heartbeatmap", "perfcounter", "rgw", "rgw_sync", "rwg_datacache", "rgw_access",
+		"rgw_dbstore", "javaclient", "asok", "throttle", "refs", "compressor", "bluestore", "bluefs", "bdev", "kstore", "rocksdb", "fuse", "mgr",
+		"mgrc", "dpdk", "eventtrace", "prioritycache", "test", "cephfs_mirror", "cepgsqlite", "seastore", "seastore_onode", "seastore_odata",
+		"seastore_ompap", "seastore_tm", "seastore_t", "seastore_cleaner", "seastore_epm", "seastore_lba", "seastore_fixedkv_tree", "seastore_cache",
+		"seastore_journal", "seastore_device", "seastore_backref", "alienstore", "mclock", "cyanstore", "ceph_exporter", "memstore", "trace"}
+)
 
 func SetLogLevel(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, storageClusterNamespace string, args []string) {
```
travisn, line 22: Let's use names for all params for readability
```suggestion
func SetLogLevel(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, storageClusterNamespace string, component, subsystem, level string) {
```

cmd/odf/set/log_level.go
```diff
@@ -11,11 +11,11 @@ import (
 
 var setLogLevelCmd = &cobra.Command{
 	Use:                "log-level",
-	Short:              "Set different log levels for ceph dameons like mon, osd and mds",
-	Example:            "odf set log-level osd 20",
+	Short:              "Set different log levels for ceph components like mon, osd and mds",
+	Example:            "odf set log-level <ceph_component> <ceph_subsystem> <{log-level}/{memory-level}>",
```
travisn, line 15: Let's just keep this as an example. The docs can give the details about the params.
```suggestion
	Example:            "odf set log-level osd crush 10",
```

travisn (changes requested):
The readme.md and set.md need updates for these changes

pkg/rook/common/logs.go
sp98: ceph will fail if invalid subsystem is passed

```
❯ bin/odf set log-level osd.0 invalid 30
Error: command terminated with exit code 22
Error EINVAL: unrecognized config option 'debug_invalid'

```

pkg/rook/common/logs.go
sp98: updated

README.md
```diff
@@ -6,7 +6,8 @@ The ODF CLI tool provides configuration and troubleshooting commands for OpenShi
 
 - `odf set recovery-profile <profile>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options `high_client_ops`, `high_recovery_ops`, or `balanced`. The default is `balanced`.
 - `odf get recovery-profile`: Get the recovery profile value.
-- `odf set log-level <ceph-daemon> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc.
+- `odf set log-level <ceph-daemon> <ceph-subsystem> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc.
```
travisn, line 9: We are going to need differentiation for setting log levels of different ODF components. Besides these ceph log levels, in the future we will want commands for the rook, noobaa, ocs-operator, or other component log levels. So I'm imagining the component as a prefix to log-level, though may need some discussion separate from this PR to finalize. And since `ceph-` is in the command name, no need for `ceph` in the param names.
```suggestion
- `odf set ceph-log-level <daemon> <subsystem> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc.
```

cmd/odf/set/log_level.go
```diff
@@ -6,44 +6,33 @@ import (
...
-
-	if !isValidSubsytem {
-		return fmt.Errorf("invalid ceph subsystem %q. Supported subsystems : %+v", args[0], common.ValidCephSubSystems)
-	}
-
-	// validate log level
-	match, err := regexp.MatchString("^[1-9][0-9]?$", args[1])
+func validateLogLevelArgs(logLevel string) error {
+	// validate {log-level}/{memory-level} format. Examples of vaild formats: 5, 1/20, 10/20, etc.
+	match, err := regexp.MatchString("^[1-9][0-9]?(\\/[1-9][0-9]?)?$", logLevel)
```
travisn, line 29: Does ceph fail if an invalid log level is set? If so, we don't even need to validate this.

pkg/rook/common/logs.go
```diff
@@ -8,11 +8,9 @@ import (
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 )
 
-var ValidCephSubSystems = []string{"osd", "mds", "mon", "mgr", "auth"}
```
travisn, line 11: Could we just move this logs.go up a directory? Seems like we don't need a `common` subdirectory. It's preferred to keep specific package names if we can avoid general names like `common` or `util`.

travisn (changes requested):

README.md
sp98: That's a really good point. Adding a prefix before log level like you mentioned sounds good.

sp98: Maybe that's where the concept of realms might have come in handy.  Something like: 
``` odf ceph set log-level <daemon> <subsystem> <log-level>
odf rook set log-level ....```

File: cmd/odf/set/log_level.go (hunk 6 above)
Comment (sp98) [reply cmd/odf/set/log_level.go:29]: Yes. Ceph does is good job in validating log levels.

``` Use "odf set [command] --help" for more information about a command.
❯ bin/odf set ceph-log-level osd.0 test 34343
Error: command terminated with exit code 34
Error ERANGE: error parsing value: value must be in range [0, 99]
❯ bin/odf set ceph-log-level osd.0 test h
Error: command terminated with exit code 22
Error EINVAL: error parsing value: value must take the form N or N/M, where N and M are integers
❯ bin/odf set ceph-log-level osd.0 test h/7
Error: command terminated with exit code 22
Error EINVAL: error parsing value: value must take the form N or N/M, where N and M are integers
❯ bin/odf set ceph-log-level osd.0 test h/k
Error: command terminated with exit code 22
Error EINVAL: error parsing value: value must take the form N or N/M, where N and M are integers

❯ bin/odf set ceph-log-level osd.0 test 4jhj

```
Although the last one is invalid, it got passed. But ceph has set the log/memory level to `4/4` for the last command. So we are good to remove this validation as well. 

Now we don't have any validation for this odf command args. Since we are adding the `debug_` prefix to ceph subsystem, I don't user can use this command just to set any config option (which was the initial concern)

README.md
sp98: now using `ceph-log-level`. Updated the README and doc file.

pkg/rook/common/logs.go
sp98: moved logs.go file to the `pkg/root` directory.

docs/set.md
```diff
@@ -21,18 +21,12 @@ odf set recovery-profile high_client_ops
...
+## ceph-log-level
 
-This command will set the log level for different subsystems. Supported subsystems:
-
-* osd
-* mds
-* mon
-* mgr
-* auth
+This command will set the log level for different ceph [subsystems](https://docs.ceph.com/en/latest/rados/troubleshooting/log-and-debug/#ceph-subsystems)
```
travisn, line 26: How about a comment like this?
```suggestion
This command will set the log level for different ceph [subsystems](https://docs.ceph.com/en/latest/rados/troubleshooting/log-and-debug/#ceph-subsystems).
The `debug_` prefix will be automatically added to the subsystem when enabling the logging.
```

pkg/rook/logs.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+)
+
+func SetCephLogLevel(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, storageClusterNamespace string, component, subsystem, level string) {
+
```
travisn, line 12: How about a log statement that indicates the setting was successfully applied? And if it fails, it already prints the error?

README.md
```diff
@@ -6,7 +6,8 @@ The ODF CLI tool provides configuration and troubleshooting commands for OpenShi
 
 - `odf set recovery-profile <profile>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options `high_client_ops`, `high_recovery_ops`, or `balanced`. The default is `balanced`.
 - `odf get recovery-profile`: Get the recovery profile value.
-- `odf set log-level <ceph-daemon> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc.
+- `odf set ceph-log-level <ceph-daemon> <ceph-subsystem> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc.
```
travisn, line 9:
```suggestion
- `odf set ceph-log-level <daemon> <subsystem> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc.
```

README.md
travisn: Still needs discussion, I could go either way on these options, but let's keep the first option for now.
```
odf set ceph-log-level ....
odf ceph set log-level ...
```
But I don't want to call it a realm, since it's really only a component name.

travisn (changes requested):

pkg/rook/logs.go
sp98: added logging statement

sp98:
``` ❯ bin/odf set ceph-log-level osd.0 crush 1/10
Info: successfully set the log levels for "osd.0" subsystem "crush" as "1/10"
```

pkg/rook/logs.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+)
+
+func SetCephLogLevel(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, storageClusterNamespace string, component, subsystem, level string) {
+
+	cephArgs := []string{"config", "set", component, fmt.Sprintf("debug_%s", subsystem), level}
+
+	exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", cephArgs, operatorNamespace, storageClusterNamespace, true, false)
```
travisn, line 16: I believe the last `false` param is for `exitOnError`. I believe we want that to be true? Or else we need to handle the return error to skip printing success.

travisn (changes requested):

pkg/rook/logs.go
sp98: yes. Missed that. Making `returnoutput` to `false` and `exitOnError` to `true`.  

``` 
❯ bin/odf set ceph-log-level osd ms 6
Info: successfully set the log levels for "osd" subsystem "ms" as "6"

❯ bin/odf set ceph-log-level osd ms y
Error EINVAL: error parsing value: value must take the form N or N/M, where N and M are integers
Error: command terminated with exit code 22

❯ bin/odf set ceph-log-level osd fgs 6
Error EINVAL: unrecognized config option 'debug_fgs'
Error: command terminated with exit code 22

❯ bin/odf set ceph-log-level osds test 6
Error EINVAL: unrecognized config target 'osds'
Error: command terminated with exit code 22
```

travisn (approved):
/approve
/lgtm

PR #24, opened 2024-02-19
mons: add command to restore mon quorum

subhamkrai:
This commits add `mons restore-quorum` command
to restore mon quorum based on single good mon.
Also adding docs for the commnad.

### DIFF
--- a/cmd/odf/mon/mon.go
+++ b/cmd/odf/mon/mon.go
@@ -1,0 +1,18 @@
+package mon
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// MonsCmd represents the mon command
+var MonsCmd = &cobra.Command{
+	Use:                "mons",
+	Short:              "Supports sub-commands for 'mons'",
+	DisableFlagParsing: true,
+	Hidden:             true,
+	Args:               cobra.ExactArgs(1),
+}
+
+func init() {
+	MonsCmd.AddCommand(RestoreQuorum)
+}
--- a/cmd/odf/mon/restore_quorum.go
+++ b/cmd/odf/mon/restore_quorum.go
@@ -1,0 +1,19 @@
+package mon
+
+import (
+	"github.com/red-hat-storage/odf-cli/cmd/odf/root"
+	"github.com/rook/kubectl-rook-ceph/pkg/mons"
+	"github.com/spf13/cobra"
+)
+
+// RestoreQuorum represents the mons command
+var RestoreQuorum = &cobra.Command{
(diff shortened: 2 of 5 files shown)


subhamkrai:
Local testing works as expected.

cmd/odf/mon/mon.go
```diff
@@ -0,0 +1,18 @@
+package mon
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// MonsCmd represents the mon command
+var MonsCmd = &cobra.Command{
+	Use:                "mons",
+	Short:              "Supports sub-commands like 'restore-quorum'",
```
travisn, line 10: I'd suggest not to mention specific subcommands, just keep it general
```suggestion
	Short:              "Supports sub-commands for mons'",
```

cmd/odf/mon/mon.go
```diff
@@ -0,0 +1,18 @@
...
+	DisableFlagParsing: true,
+	Hidden:             true,
```
travisn, line 12: All of the mon commands are hidden? I was thinking only advanced commands like restore-quorum might need to be hidden.

cmd/odf/mon/restore_quorum.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/spf13/cobra"
+)
+
+// RestoreQuorum represents the mons command
+var RestoreQuorum = &cobra.Command{
+	Use:                "restore-quorum",
+	Short:              "When quorum is lost, restore quorum to the remaining healthy mon",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Example:            "odf mons restore-quorum <Mon_ID>",
```
travisn, line 15: nit: CLI examples are better lower-case IMO
```suggestion
	Example:            "odf mons restore-quorum <mon_id>",
```

travisn (changes requested):
a few small suggestions

cmd/odf/mon/mon.go
subhamkrai: got it

cmd/odf/mon/mon.go
subhamkrai: IIRC, Eran's comment was to have all the dev preview command hidden. I guess it is also mentioned in Jira by him

travisn: Let's discuss this more. I see the point of hiding dev preview items, but also would prefer not to.

travisn (approved):

PR #26, opened 2024-02-28
build: update kubectl-rook-ceph version

subhamkrai:
updating kubectl-rook-ceph to v0.8.0 latest version.


travisn (approved):
/approve
/lgtm

PR #27, opened 2024-02-28
rook: add command rook status

subhamkrai:
rook: add command rook status
add command rook status for printing rook-ceph CR status.

### DIFF
--- a/cmd/odf/get/rook_status.go
+++ b/cmd/odf/get/rook_status.go
@@ -1,0 +1,27 @@
+package get
+
+import (
+	"github.com/red-hat-storage/odf-cli/cmd/odf/root"
+	"github.com/rook/kubectl-rook-ceph/pkg/rook"
+	"github.com/spf13/cobra"
+)
+
+var rookCmd = &cobra.Command{
+	Use:   "rook",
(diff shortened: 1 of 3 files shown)


cmd/odf/get/rook_status.go
```diff
@@ -0,0 +1,28 @@
...
+var rookCmd = &cobra.Command{
+	Use:    "rook",
+	Short:  "Supports sub-commands for 'rook'",
+	Args:   cobra.RangeArgs(1, 2),
+	Hidden: true,
+}
+
+var crStatusCmd = &cobra.Command{
+	Use:     "status",
+	Short:   "Print the phase and/or conditions of every CR in the namespace",
```
travisn, line 18:
```suggestion
	Short:   "Print the phase and/or conditions of Rook CRs",
```

travisn (changes requested):
After this is rebased, just a minor suggestion...

cmd/odf/get/rook_status.go
subhamkrai: done

travisn (approved):

PR #31, opened 2024-03-06
restore: add command to restore crd

subhamkrai:
This commits add command restore crd which are stuck in terminating state.


subhamkrai:
@travisn 

I was thinking to have a new command `restore`  and it will have sub-commands like
```bash

- restore
  -  crds
  - mon-quorum
```

currently, to restore mon-quorum the command is `odf mons restore-quorum <mon_id>` but I like the first approach as it will be more easier for the user to find a command with `restore`. WDYT?

cmd/odf/restore/crds.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/spf13/cobra"
+)
+
+// crdsCmd represents the mons command
+var crdsCmd = &cobra.Command{
+	Use:                "crds",
+	Short:              "Restores a CR that was accidentally deleted and is still in terminating state.",
+	DisableFlagParsing: true,
+	Args:               cobra.RangeArgs(1, 2),
+	Example:            "odf restore crds <CRD> [CRNAME]",
```
travisn, line 16: What about the command `deleted`? IMO it's important to be clear in the command name that the CRs to be restored are in the deleted state.
```suggestion
	Example:            "odf restore deleted <CRD> [CRNAME]",
```

cmd/odf/restore/restore.go
```diff
@@ -0,0 +1,16 @@
+package restore
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// restoreCrd represents the mons command
+var RestoreCrd = &cobra.Command{
+	Use:                "restore",
+	Short:              "Calls subcommands for 'restore'",
```
travisn, line 10: I wonder if we don't need a short description for the higher level commands like this? It just seems obvious already that subcommands will be implemented.

travisn (changes requested):

cmd/odf/restore/restore.go
subhamkrai: I'll remove in separate PR for other commands

cmd/odf/restore/crds.go
subhamkrai: agreed, updated the command.

travisn (approved):

PR #34, opened 2024-03-26
update kubectl-rook-ceph to include latest changes

yati1998:
This commit updates kubectl-rook-ceph to
include the lateste changes


subhamkrai (commented):
@yati1998 we don't need any extra changes as the commands and function calls are thes eame just updating the version will work?

yati1998:
> @yati1998 we don't need any extra changes as the commands and function calls are thes eame just updating the version will work?

No, we don't need any additional cde change here, I have tested it to confirm as well. It works fine.

subhamkrai (approved):
lgtm

PR #40, opened 2024-04-25
osd: add command to set osd full ratio

subhamkrai:


cmd/odf/set/osd_settings.go
```diff
@@ -0,0 +1,52 @@
...
+		cephArgs := []string{"osd", "set-full-ratio", args[0]}
+		out, err := exec.RunCommandInOperatorPod(ctx, root.ClientSets, "ceph", cephArgs, root.OperatorNamespace, root.StorageClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(fmt.Errorf("failed to run command ceph with args %v: %v", cephArgs, err))
+		}
+		logging.Info(out)
+	},
+}
+
+func validateArgs(arg string) error {
```
travisn, line 36: What error does ceph give if an invalid is passed and we don't validate it? Perhaps ceph's checks are sufficient and we don't need to validate?

cmd/odf/set/osd_settings.go
```diff
@@ -0,0 +1,52 @@
...
+	checkIfFloat := strings.Contains(arg, ".")
```
travisn, line 37: Is "0" or "1" not a valid input for ceph?

cmd/odf/set/osd_settings.go
```diff
@@ -0,0 +1,52 @@
...
+	if !checkIfFloat {
+		return fmt.Errorf("arg should be a float and must be in range of '0.0' to '1.0'")
+	}
+
+	val, err := strconv.Atoi(arg)
```
travisn, line 42: I think you want `strconv.ParseFloat()`

travisn (changes requested):

cmd/odf/set/osd_settings.go
subhamkrai: right we don't need to validate
```bash
./bin/odf -n rook-ceph set ceph osd set-full-ratio 2.0
Error: failed to run command ceph with args [osd set-full-ratio 2.0]: Invalid command: 2.0 not in range [0.0, 1.0]
osd set-full-ratio <ratio:float> :  set usage ratio at which OSDs are marked full
Error EINVAL: invalid command
. failed to run command. command terminated with exit code 22
```

cmd/odf/set/osd_settings.go
subhamkrai: no required any more with latest changes

cmd/odf/set/osd_settings.go
subhamkrai: no required any more with latest changes

travisn (changes requested):
Will there be a separate PR for setting the near-full ratio and the backfill ratio? You might consider including them in the same PR so we can review for consistency of the commands.

cmd/odf/set/osd_settings.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+
+	"github.com/red-hat-storage/odf-cli/cmd/odf/root"
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/spf13/cobra"
+)
+
+var osdFullRatioCmd = &cobra.Command{
+	Use:                "set-full-ratio",
```
travisn, line 13: How about a more general approach for the `set` commands? The command could be: 
```
odf set ceph <key> <value>
```

The key must be from a supported list, so we don't allow any ceph setting. This way it is much less code to add more keys that we want to allow setting.

cmd/odf/set/osd_settings.go
```diff
@@ -0,0 +1,46 @@
...
+		cephArgs := []string{"osd", "set-full-ratio", args[0]}
+		out, err := exec.RunCommandInOperatorPod(ctx, root.ClientSets, "ceph", cephArgs, root.OperatorNamespace, root.StorageClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(fmt.Errorf("failed to run command ceph with args %v: %v", cephArgs, err))
+		}
+		logging.Info(out)
+	},
+}
+
+var osdBackFillFulRatioCmd = &cobra.Command{
```
travisn, line 30: There is also a `nearfull` setting

travisn:
> > Will there be a separate PR for setting the near-full ratio and the backfill ratio? You might consider including them in the same PR so we can review for consistency of the commands.
> 
> added now, also two question:
> 
> 1. currently after running the commands it prints blank  line like below, should we add some logs when it is applied
> 
> ```shell
> ./bin/odf -n rook-ceph set ceph osd set-full-ratio 0.9
> 
> ~/go/src/github.com/odf-cli
> srai@fedora ~ (add-cmd-osd-full-ratio) $ 
> ```

Sounds good to print a success message, such as `Successfully set <key> to <value>`

> 2. should add `get` command similar to set for these two new commands to list the settings? Also, we could call the `get` command after `set` command which prints the config. wdyt?

Yes, `get` commands sound good. How about in a follow-up PR? But perhaps we don't call it as part of the `get` since with the suggested generic approach, it might not work for all commands.

travisn (changes requested):

subhamkrai:
> > > Will there be a separate PR for setting the near-full ratio and the backfill ratio? You might consider including them in the same PR so we can review for consistency of the commands.
> > 
> > 
> > added now, also two question:
> > 
> > 1. currently after running the commands it prints blank  line like below, should we add some logs when it is applied
> > 
> > ```shell
> > ./bin/odf -n rook-ceph set ceph osd set-full-ratio 0.9
> > 
> > ~/go/src/github.com/odf-cli
> > srai@fedora ~ (add-cmd-osd-full-ratio) $ 
> > ```
> 
> Sounds good to print a success message, such as `Successfully set <key> to <value>`
> 
> > 2. should add `get` command similar to set for these two new commands to list the settings? Also, we could call the `get` command after `set` command which prints the config. wdyt?
> 
> Yes, `get` commands sound good. How about in a follow-up PR? But perhaps we don't call it as part of the `get` since with the suggested generic approach, it might not work for all commands.

follow up pr sounds good

cmd/odf/set/osd_settings.go
subhamkrai: added

cmd/odf/set/osd_settings.go
subhamkrai: I made the changes but there is one more command that sets the recover profile as syntax
```
odf set recovery-profile <option>
```

may be we can move this command also under 'ceph' sub-command? as follow up pr?

docs/set.md
```diff
@@ -35,3 +36,35 @@ odf set ceph log-level osd crush 20
...
+### osd
+
+#### full-ratio
+
+This command allows updating the ceph osd full ratio in case ceph prevents the I/O operation on OSDs that reached the capacity specified.
+
+**Note**: If the value is set too close to 1.0, the cluster will be unrecoverable if the OSDs are full and there is nowhere to grow.
+
+``` bash
+odf set ceph full-ratio 0.9
```
travisn, line 49: Now I'm actually wondering if we should have these as higher level settings instead of under ceph, let's discuss. There shouldn't be any conflict with other components setting a similar setting.
```suggestion
odf set full-ratio 0.9
```

subhamkrai: ass discussed moving it top-level command

README.md
```diff
@@ -6,7 +6,11 @@ The ODF CLI tool provides configuration and troubleshooting commands for OpenShi
 
 - `odf set`:
   - `recovery-profile <profile>`: Set the recovery profile to favor new IO, recovery, or balanced mode with options `high_client_ops`, `high_recovery_ops`, or `balanced`. The default is `balanced`.
-  - `ceph log-level <daemon> <subsystem> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc. More information about the ceph subsystems can be found [here](https://docs.ceph.com/en/latest/rados/troubleshooting/log-and-debug/#ceph-subsystems)
+  - `ceph`
+    - `log-level <daemon> <subsystem> <log-level>`: Set the log level for Ceph daemons like OSD, mon, mds etc. More information about the ceph subsystems can be found [here](https://docs.ceph.com/en/latest/rados/troubleshooting/log-and-debug/#ceph-subsystems)
+    - `full-ratio <value>`: Update the ceph osd full ration after which Ceph automatically prevents any I/O operations on OSDs.
```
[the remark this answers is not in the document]
subhamkrai: corrected

cmd/odf/set/backfillfull_ratio.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/spf13/cobra"
+)
+
+var osdBackFillFullRatioCmd = &cobra.Command{
+	Use:                "backfillfull",
+	Short:              "Configure ceph osd 'backfillfull-ratio' setting",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Example:            "odf set ceph backfillfull-ratio 0.9",
```
travisn, line 17: For all the commands:
```suggestion
	Example:            "odf set backfillfull-ratio 0.9",
```

docs/set.md
```diff
@@ -22,7 +25,38 @@ odf set recovery-profile high_client_ops
 
 To verify the recovery profile setting run [odf get recovery-profile](get.md#recovery-profile).
 
+## full-ratio
```
travisn, line 28: match the command name in the titles?
```suggestion
## full
```

travisn (changes requested):
a few nits

docs/set.md
subhamkrai: done

cmd/odf/set/backfillfull_ratio.go
subhamkrai: done

docs/set.md
```diff
@@ -22,7 +25,38 @@ odf set recovery-profile high_client_ops
 
 To verify the recovery profile setting run [odf get recovery-profile](get.md#recovery-profile).
 
+## full
```
travisn, line 28: One more request... It would be helpful in the examples to show ratios that are likely to be used, as well as document the defaults that we have in ODF. So if the defaults are .75/.80/.85 for nearfull/backfillfull/full, then how about we use example ratios of 0.05 higher at .80/.85/.90?

subhamkrai: done

travisn (approved):

PR #41, opened 2024-04-29
ci: add dependatbot to update deps

subhamkrai:
Please check individual commits

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    groups:
+      golang-dependencies:
+        patterns:
+          - "github.com/golang*"
+          - "golang.org/x/*"
+      k8s-dependencies:
+        patterns:
+          - "k8s.io*"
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+<!-- Thank you for contributing to Rook! -->
```
travisn, line 1: This template needs to be customized for odf, instead of having all the rook references

.github/workflows/commitlint.yaml
```diff
@@ -0,0 +1,37 @@
+name: Commitlint
```
travisn, line 1: Is there a commit linter for ocs operator or other odf repos? We may not have the same checks, but for all the downstream-only repos like this, hopefully prow can give the consistency, so maybe we don't need this commit linter?

travisn (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
subhamkrai: I removed this commit also it doesn't make sense with commitlint commit to me.

.github/workflows/commitlint.yaml
subhamkrai: yeah removed this commit

travisn (approved):

PR #47, opened 2024-05-15
core: remove hidden true from all cmd

subhamkrai:
since with 4.17, all the commands will GA'ed removing `Hidden: true` which hides the command from the help section.


subhamkrai (commented):
@travisn do we want these changes in-order to GA all the commands?

travisn:
> @travisn do we want these changes in-order to GA all the commands?

Yes I believe so, but let's discuss.

PR #53, opened 2024-06-04
Add new commands for DR pre-requisite

umangachapagain:


travisn (commented):
Please also define the command(s) in the main readme.md, and the subtopics such as set.md. This will help understand the commands being reviewed, thanks!

docs/set.md
```diff
@@ -70,4 +71,12 @@ odf set ceph log-level osd crush 20
...
 Once the logging efforts are complete, restore the systems to their default or to a level suitable for normal operations.
 
+## multiclusterservice
 
+This command is used to set the MultiClusterService spec on StorageCluster resource. It takes a ManagedClusterName as argument.
+
+```bash
+$ odf set multiclusterservice peer-cluster-1
+
+Info: Successfully set 'multiClusterService' on StorageCluster "ocs-storagecluster" in namespace "openshift-storage".
```
travisn, line 81: This command simply sets the property on the StorageCluster CR? The same thing could be done with an `oc` command, so I would hesitate to have it in the ODF CLI unless there is a very clear need. It it only to simplify the CLI command compared to an `oc` command to patch the StorageCluster CR?

umangachapagain: Removed this command.

cmd/odf/get/dr_prereq.go
```diff
@@ -0,0 +1,109 @@
...
+
+const submarinerOperatorNamespace string = "submariner-operator"
+
+var drPrereqCmd = &cobra.Command{
+	Use:                "dr-prereq",
+	Short:              "Print the status of pre-requisites for Disaster Recovery between peer clusters.",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Example:            "odf get dr-prereq <peer-clusterID>",
+	Run: func(cmd *cobra.Command, args []string) {
```
subhamkrai, line 28: @umangachapagain can we move implementation details of `Run:` to under `pkg/` folder to make `cmd/` more clean? Thanks

subhamkrai (changes requested):

cmd/odf/get/dr_prereq.go
umangachapagain: Moved.

cmd/odf/get/dr_prereq.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/red-hat-storage/odf-cli/pkg/odf/dr"
+	"github.com/spf13/cobra"
+)
+
+var drPrereqCmd = &cobra.Command{
+	Use:                "dr-prereq",
+	Short:              "Print the status of pre-requisites for Disaster Recovery between peer clusters.",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Example:            "odf get dr-prereq <peer-clusterID>",
```
travisn, line 13: In the readme the arg is called a cluster name, rather than an ID. Should one be renamed for consistency?

pkg/odf/dr/dr.go
```diff
@@ -0,0 +1,98 @@
...
+	submarinerv1alpha1 "github.com/submariner-io/submariner-operator/api/v1alpha1"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	"github.com/submariner-io/submariner/pkg/cidr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const submarinerOperatorNamespace string = "submariner-operator"
```
travisn, line 19: The submariner operator is always in this ns?

pkg/odf/dr/dr.go
```diff
@@ -0,0 +1,98 @@
...
+
+	peerClusterCR := &submarinerv1.Cluster{}
+	err = client.Get(ctx, types.NamespacedName{Name: peerClusterID, Namespace: submarinerOperatorNamespace}, peerClusterCR)
+	if err != nil {
+		return false, err
+	}
+
+	err = cidr.OverlappingSubnets(clusterCR.Spec.ServiceCIDR, clusterCR.Spec.ClusterCIDR,
+		append(peerClusterCR.Spec.ClusterCIDR, peerClusterCR.Spec.ServiceCIDR...))
+	if err != nil {
```
travisn, line 86: Every failure for overlapping subnets just indicates globalnet is required? No other kind of failure?

travisn (changes requested):
Just a few questions or nits

pkg/odf/dr/dr.go
umangachapagain: Yes

cmd/odf/get/dr_prereq.go
umangachapagain: Agreed. Changing this to `<PeerManagedClusterName>`

pkg/odf/dr/dr.go
umangachapagain: Yes. Error is only returned when overlapping subnets are found.

cmd/odf/root/root.go
```diff
@@ -92,6 +110,16 @@ func getClientsets(ctx context.Context) *k8sutil.Clientsets {
...
 
+func getControllerRuntimeClient() ctrl.Client {
+	client, err := ctrl.New(ClientSets.KubeConfig, ctrl.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		logging.Fatal(err)
+	}
+	return client
+}
```
subhamkrai, line 121:
```suggestion
	client, err := ctrl.New(ClientSets.KubeConfig, ctrl.Options{
		Scheme: scheme,
	})
	if err != nil {
		logging.Fatal(err)
	}
	return client
}
```

@umangachapagain can we move this implementation inside `getClientsets` similar to other clients?

pkg/odf/dr/dr.go
subhamkrai: do we want to cli option for this to overwrite?

subhamkrai (changes requested):
small nits, overall looks good

travisn (approved):
LGTM after Subham's final questions are answered

travisn:
Ugh, the bot merged it after a single approval even though there were still outstanding requests. I guess we'll need a follow-up PR, thanks.

pkg/odf/dr/dr.go
umangachapagain: I don't see the need as this is hardcoded within submariner project as well.
We can assess it when the need arises.

cmd/odf/root/root.go
umangachapagain: controller-runtime clients are implemented differently from other ClientSets. We could even replace all the clients in the ClientSets with this single client. That is why I chose to keep it separate (it is not a ClientSet type).

I also looked at replacing other clientsets with this client to keep the dependencies minimal and easy to manage. But, the refactor was needed in the rook-ceph CLI as well (due to hard dependency between the projects). Given controller-runtime client is recommended over ClientSets, should we do a refactor in that direction? @travisn @subhamkrai

subhamkrai: > controller-runtime clients are implemented differently from other ClientSets. We could even replace all the clients in the ClientSets with this single client. That is why I chose to keep it separate (it is not a ClientSet type).
> 
> I also looked at replacing other clientsets with this client to keep the dependencies minimal and easy to manage. But, the refactor was needed in the rook-ceph CLI as well (due to hard dependency between the projects). Given controller-runtime client is recommended over ClientSets, should we do a refactor in that direction? @travisn @subhamkrai

@umangachapagain yes, sounds good to do the refactoring.

subhamkrai: Will you open a follow-up PR for this?

umangachapagain: Sure. Might take a while, but will do.

subhamkrai: sure, Thanks

PR #74, opened 2024-08-14
Bug 2294118: build:  bump github.com/rook/rook from 1.14.7 to 1.14.8

subhamkrai:


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.21
 require (
 	github.com/pkg/errors v0.9.1
 	github.com/rook/kubectl-rook-ceph v0.8.1-0.20240327033245-3f07e9fd3059
-	github.com/rook/rook v1.13.7
+	github.com/rook/rook v1.14.8
```
travisn, line 8: Any reason not to bump it to v1.14.9?

travisn: Nevermind, this is just a backport, let's keep that

PR #105, opened 2025-01-06
build: update go 1.23 and rook 1.16.1

subhamkrai:
updating go version to 1.23 and toolcain to 1.23.4 with other updates.


travisn (approved):
/approve
/lgtm

PR #114, opened 2025-02-04
Add Support for Restoring OCS Operator CRs in ODF CLI

OdedViner:
This PR adds support for restoring deleted Custom Resources (CRs) for the OCS operator via the ODF CLI. It includes logic for handling storageclusters CRs by dynamically setting groupName and versionResource, ensuring compatibility and reliability during the restore process.


cmd/odf/restore/crds.go
```diff
@@ -20,7 +21,16 @@ var deletedCmd = &cobra.Command{
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		k8sutil.SetDeploymentScale(cmd.Context(), root.ClientSets.Kube, root.OperatorNamespace, "ocs-operator", 0)
-		pkgrestore.RestoreCrd(cmd.Context(), root.ClientSets, root.OperatorNamespace, root.StorageClusterNamespace, args)
+		var groupName string
+		var versionResource string
+		if args[0] == "storageclusters" {
```
travisn, line 26: Is there a check for length of `args` before this? If not, this could crash.

travisn (changes requested):
With this change, we really should enable the restore of any of the ODF CRs, whether from Rook, OCS operator, CSI operator, Noobaa, etc. 

One approach would be to just require `args[0]` to include the fully qualified CRD type such as `storageclusters.ocs.openshift.io`, then we just extract the group name from that full name. We could return an error if they passed some group name that we don't recognize as belonging to ODF.

cmd/odf/restore/crds.go
OdedViner: The command already includes argument validation through Cobra with the line:
```
Args: cobra.RangeArgs(1, 2),
```

This ensures that if the number of arguments is outside the 1–2 range, Cobra will return an error and prevent further execution. As shown in the output, passing three arguments results in:

```
$ ./bin/odf restore deleted  a b c
Error: accepts between 1 and 2 arg(s), received 3
```

OdedViner:
> With this change, we really should enable the restore of any of the ODF CRs, whether from Rook, OCS operator, CSI operator, Noobaa, etc.
> 
> One approach would be to just require `args[0]` to include the fully qualified CRD type such as `storageclusters.ocs.openshift.io`, then we just extract the group name from that full name. We could return an error if they passed some group name that we don't recognize as belonging to ODF.

Hi @travisn ,

I’ve implemented the change as suggested. The restore command now requires that the first argument be a fully qualified CRD type (for example, storageclusters.ocs.openshift.io). The code parses this input to extract the resource name, group, and API version, and it validates the group against a set of supported groups (Rook, OCS, CSI, NooBaa, etc.). If the provided group isn’t recognized as part of ODF, the command returns an error with a helpful message listing the supported groups.

This update should enable us to restore any of the ODF CRs regardless of the operator to which they belong. Please take a look at the changes and let me know if you have any further suggestions or feedback.

Thanks,
Oded

subhamkrai (changes requested):

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,53 @@
...
 )
 
+// groupVersions defines the supported CRD groups and their corresponding API versions.
+var groupVersions = map[string]string{
+	"ocs.openshift.io":       "v1",
+	"ceph.rook.io":           "v1",
+	"storage.k8s.io":         "v1",
+	"odf.openshift.io":       "v1alpha1",
+	"noobaa.io":              "v1alpha1",
+	"csiaddons.openshift.io": "v1alpha1",
```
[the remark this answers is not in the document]
subhamkrai: it is v1alpha1 please confirm and upate

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,54 @@
...
+		out = append(out, k)
+	}
+	return out
+}
+
+// parseFullyQualifiedCRD takes a fully qualified CRD type of the form "resource.group"
+// (for example, "cephclusters.ceph.rook.io") and returns the resource name, group name, and
+// the API version associated with that group. It returns an error if the format is invalid
+// or the group is not recognized.
+func parseFullyQualifiedCRD(fqcrd string) (resourceName, groupName, version string, err error) {
```
travisn, line 37: Please add a unit test for this method. Unit tests are always valuable for parsing.

cmd/odf/restore/crds.go
```diff
@@ -20,8 +62,19 @@ var deletedCmd = &cobra.Command{
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		k8sutil.SetDeploymentScale(cmd.Context(), root.ClientSets.Kube, root.OperatorNamespace, "ocs-operator", 0)
-		pkgrestore.RestoreCrd(cmd.Context(), root.ClientSets, root.OperatorNamespace, root.StorageClusterNamespace, args)
+		// Parse the fully qualified CRD (e.g. "cephclusters.ceph.rook.io").
+		resourceName, groupName, version, err := parseFullyQualifiedCRD(args[0])
+		if err != nil {
+			fmt.Printf("Error parsing CRD type: %v\n", err)
```
travisn, line 68: Instead of printing the error, see the helper for `logging.Fatal()`

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,54 @@
...
 	pkgrestore "github.com/rook/kubectl-rook-ceph/pkg/restore"
 	"github.com/spf13/cobra"
 )
 
+// groupVersions defines the supported CRD groups and their corresponding API versions.
+var groupVersions = map[string]string{
+	"ocs.openshift.io":       "v1",
+	"ceph.rook.io":           "v1",
+	"storage.k8s.io":         "v1",
+	"csi.ceph.io":            "v1",
```
travisn, line 18: In 4.19 we are pushing for v1beta1 for the csi CRDs, let's assume that for now
```suggestion
	"csi.ceph.io":            "v1beta1",
```

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,54 @@
...
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 	pkgrestore "github.com/rook/kubectl-rook-ceph/pkg/restore"
 	"github.com/spf13/cobra"
 )
 
+// groupVersions defines the supported CRD groups and their corresponding API versions.
+var groupVersions = map[string]string{
+	"ocs.openshift.io":       "v1",
+	"ceph.rook.io":           "v1",
+	"storage.k8s.io":         "v1",
```
travisn, line 17: Is this for csidrivers and storageclasses? Seems like we shouldn't need to restore this type. ODF might create resources of that type, but I don't anticipate they would have finalizers that we would need to worry about.

travisn (changes requested):

cmd/odf/restore/crds.go
OdedViner: done

cmd/odf/restore/crds.go
OdedViner: done

cmd/odf/restore/crds.go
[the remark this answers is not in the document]
OdedViner: done

cmd/odf/restore/crds.go
OdedViner: removed "storage.k8s.io" CRD

OdedViner:
This PR depends on the kubectl-rook-ceph PR https://github.com/rook/kubectl-rook-ceph/pull/354 that was merged today.
@subhamkrai @travisn , is there an option to run a bot to update the dependencies?

subhamkrai:
> This PR depends on the kubectl-rook-ceph PR [rook/kubectl-rook-ceph#354](https://github.com/rook/kubectl-rook-ceph/pull/354) that was merged today. @subhamkrai @travisn , is there an option to run a bot to update the dependencies?

That will require kubectl-rook-ceph plugin release which I'm not sure we are planning now but you can use kubectl-rook-ceph with commit that has your changes

OdedViner:
> > This PR depends on the kubectl-rook-ceph PR [rook/kubectl-rook-ceph#354](https://github.com/rook/kubectl-rook-ceph/pull/354) that was merged today. @subhamkrai @travisn , is there an option to run a bot to update the dependencies?
> 
> That will require kubectl-rook-ceph plugin release which I'm not sure we are planning now but you can use kubectl-rook-ceph with commit that has your changes

done

subhamkrai (commented):
Changes looks @OdedViner have you verified the changes are working as expected.

OdedViner:
When attempting to restore it storagecluster CR, the process encountered an error while trying to remove finalizers from the storageclusters CR
```
1.Get storagecluster
$ oc get storagecluster -n openshift-storage 
NAME                 AGE   PHASE   EXTERNAL   CREATED AT             VERSION
ocs-storagecluster   21h   Ready              2025-02-23T18:25:32Z   4.18.0

2.Delete Storagecluster
$ oc delete storagecluster ocs-storagecluster -n openshift-storage 
storagecluster.ocs.openshift.io "ocs-storagecluster" deleted

3.Verify storagecluster in deleteing state
$ oc get storagecluster -n openshift-storage 
NAME                 AGE   PHASE      EXTERNAL   CREATED AT             VERSION
ocs-storagecluster   21h   Deleting              2025-02-23T18:25:32Z   4.18.0

4.Run Restore CLI 
$ oc get crd | grep storageclusters
storageclusters.ocs.openshift.io                                  2025-02-23T18:19:26Z

$ ./bin/odf restore deleted storageclusters.ocs.openshift.io
Info: Detecting which resources to restore for crd "storageclusters"
Info: Restoring CR ocs-storagecluster
Warning: The resource ocs-storagecluster was found deleted. Do you want to restore it? yes | no

yes
Info: Proceeding with restoring deleting CR
Info: Scaling down the operator
Info: Deleting validating webhook rook-ceph-webhook if present
Info: Removing ownerreferences from resources with matching uid ad51048e-cad3-47ce-88a7-c599cd14869a
Info: Removing owner references for secret onboarding-private-key
Info: Removed ownerReference for Secret: onboarding-private-key

Info: Removing owner references for secret onboarding-ticket-key
Info: Removed ownerReference for Secret: onboarding-ticket-key

Info: Removing owner references for configmaps rook-config-override
Info: Removed ownerReference for configmap: rook-config-override

Info: Removing owner references for service ocs-metrics-exporter
Info: Removed ownerReference for service: ocs-metrics-exporter

Info: Removing owner references for deployment ocs-metrics-exporter
Info: Removed ownerReference for deployment: ocs-metrics-exporter

Info: Removing owner references for deployment rook-ceph-tools
Info: Removed ownerReference for deployment: rook-ceph-tools

Info: Removing finalizers from storageclusters/ocs-storagecluster
Error: Failed to update resource "ocs-storagecluster" for crd. the server could not find the requested resource
```

travisn:
@OdedViner How's the testing looking for this? It would be good to have this work completed soon.

subhamkrai (commented):
Please squah your commits and did you tested the latest changes with odf cluster?

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,63 @@
...
+	"ocs.openshift.io":       "v1",
+	"ceph.rook.io":           "v1",
+	"csi.ceph.io":            "v1beta1",
+	"odf.openshift.io":       "v1alpha1",
+	"noobaa.io":              "v1alpha1",
+	"csiaddons.openshift.io": "v1alpha1",
+}
+
+// keys returns the keys of a string map. It is used to print out supported group names.
+func keys(m map[string]string) []string {
```
subhamkrai, line 25:
```suggestion
func groupNameKeys(m map[string]string) []string {
```
Let's have name related to their work

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,63 @@
...
+// or the group is not recognized.
+func parseFullyQualifiedCRD(fqcrd string) (resourceName, groupName, version string, err error) {
+	parts := strings.SplitN(fqcrd, ".", 2)
+	if len(parts) != 2 {
+		return "", "", "", fmt.Errorf("invalid CRD format %q; expected format <resource>.<group>", fqcrd)
+	}
+	resourceName = parts[0]
+	groupName = parts[1]
+
+	ver, ok := groupVersions[groupName]
```
subhamkrai, line 45:
```suggestion
	version, ok := groupVersions[groupName]
```

cmd/odf/restore/crds.go
```diff
@@ -20,8 +71,54 @@ var deletedCmd = &cobra.Command{
...
+			logging.Fatal(fmt.Errorf("Error parsing CRD type: %v\n", err))
+		}
+		// Construct a new args slice with the resource name as the first argument.
+		newArgs := make([]string, len(args))
+		newArgs[0] = resourceName
+		if len(args) > 1 {
+			newArgs[1] = args[1]
+		}
+		var customResources []pkgrestore.CustomResource
+		if contains(newArgs, "storageclusters") {
```
subhamkrai, line 86:
```suggestion
		if contains(newArgs, "storageclusters") {
```
I think we can use `string.Slices` here without https://go.dev/play/p/sj50jslDBUY to make code clean

subhamkrai (changes requested):
Apart from these nits, changes looks good to me

cmd/odf/restore/crds.go
OdedViner: done

cmd/odf/restore/crds.go
OdedViner: done

cmd/odf/restore/crds.go
OdedViner: done

subhamkrai:
The changes look good to me. Thanks

@travisn, ready to merge?
If I approve, the bot will merge, so I have not put lgtm tag

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,55 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	pkgrestore "github.com/rook/kubectl-rook-ceph/pkg/restore"
 	"github.com/spf13/cobra"
 )
 
+// groupVersions defines the supported CRD groups and their corresponding API versions.
+var groupVersions = map[string]string{
+	"ocs.openshift.io":       "v1",
+	"ceph.rook.io":           "v1",
+	"csi.ceph.io":            "v1beta1",
```
travisn, line 19: In 4.19 we decided to keep this as v1alpha1, then in 4.20 it will change to `v1`.
```suggestion
	"csi.ceph.io":            "v1alpha1",
```

go.sum
```diff
@@ -743,8 +743,8 @@ github.com/rogpeppe/go-internal v1.6.1/go.mod h1:xXDCJY+GAPziupqXw64V24skbSoqbTE
 github.com/rogpeppe/go-internal v1.8.1/go.mod h1:JeRgkft04UBgHMgCIwADu4Pn6Mtm5d4nPKWu0nJ5d+o=
 github.com/rogpeppe/go-internal v1.12.0 h1:exVL4IDcn6na9z1rAb56Vxr+CgyK3nn3O+epU5NdKM8=
 github.com/rogpeppe/go-internal v1.12.0/go.mod h1:E+RYuTGaKKdloAfM02xzb0FW3Paa99yedzYV+kq4uf4=
-github.com/rook/kubectl-rook-ceph v0.9.3 h1:+7THA8a+S2ArJrs9jpY1eJscAjPBKjlLXSmCVPU3eoY=
-github.com/rook/kubectl-rook-ceph v0.9.3/go.mod h1:dOQ+Yccc41DxZqe9jpvAUHsYTquYP/SKClrPmG70SLM=
+github.com/rook/kubectl-rook-ceph v0.9.4-0.20250428051344-dbfe77cc57a1 h1:znvPe0apxkTdkdiVINk0DfUbMGt6Vv+I9mgKfsr3odY=
```
travisn, line 746: Discussed with @subhamkrai today that he will plan on v0.9.4 tomorrow, if you prefer to update this reference after that release.

travisn (changes requested):

go.sum
subhamkrai: https://github.com/rook/kubectl-rook-ceph/releases/tag/v0.9.4

cmd/odf/restore/crds.go
OdedViner: done

go.sum
OdedViner: done

travisn (commented):
LGTM, it seems we just need to resolve the questions in https://issues.redhat.com/browse/DFBUGS-793 before merging.

cmd/odf/restore/crds.go
```diff
@@ -1,12 +1,55 @@
...
+	resourceName = parts[0]
+	groupName = parts[1]
+
+	version, ok := groupVersions[groupName]
+	if !ok {
+		return "", "", "", fmt.Errorf("unsupported group %q; supported groups are: %v", groupName, groupNameKeys(groupVersions))
+	}
+	return resourceName, groupName, version, nil
+}
+
```
leelavg, line 52: any reason in not using the established GVK/GVR schema from https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/schema/group_version.go or https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/group_version.go if we want a higher level of abstraction.

is it like you don't want to import Go Structs from other repos? as of now iiuc, we already made an effort moving APIs into a submodule.

cmd/odf/restore/crds.go
```diff
@@ -20,8 +63,54 @@ var deletedCmd = &cobra.Command{
...
+		newArgs := make([]string, len(args))
+		newArgs[0] = resourceName
+		if len(args) > 1 {
+			newArgs[1] = args[1]
+		}
+		var customResources []pkgrestore.CustomResource
+		if slices.Contains(newArgs, "storageclusters") {
+			customResources = []pkgrestore.CustomResource{
+				// ceph.rook.io/v1
+				{Group: "ceph.rook.io", Version: "v1", Resource: "cephblockpoolradosnamespaces"},
```
leelavg, line 81: how can we create a link back to the ocs-op to keep this list updated when a new resource is owned by storagecluster?

did you check that k8s primitives (secrets, configmaps etc) that are owned by storagecluster are ok to be deleted?

apart from ensuring pods are running what other checks are performed after restoration of storagecluster CR?

iirc, storagecluster UID is being used in DR related activities as well, maybe that also gets effected or we need to have a statement of not performing this on DR enabled setups?

cmd/odf/restore/crds.go
OdedViner: @leelavg The input format is `./bin/odf restore deleted storageclusters.ocs.openshift.io`, where only the CR name and group are provided (storageclusters.ocs.openshift.io). We do not specify the version. Please add the relevant code to improve the current implementation.

cmd/odf/restore/crds.go
OdedViner: @leelavg I have not tested it on a DR setup or in Provider mode.

OdedViner:
> just dropping it here as I didn't look into the plugins that we were developing upto now (pardon), usually https://github.com/kubernetes/cli-runtime is preferred as a base when developing kubectl plugins as it abstracts api-resources, discovery if required, generic printers etc.

@leelavg This is a downstream project, so I believe it would be more relevant for the upstream project: https://github.com/rook/kubectl-rook-ceph.

subhamkrai:
We also need to scale down `odf-operator` when restoring storageCluster.

OdedViner:
> We also need to scale down `odf-operator` when restoring storageCluster.

@subhamkrai 
The `StorageCluster` is an OCS CR, and I don't think the odf-operator is relevant here. As I understand it, the `odf-operator` is responsible for installing all operators, including OCS, Rook, NooBaa, Ceph CS and etc..

There is also a `StorageSystem` CR managed by the `odf-operator`, which allows for deploying multiple StorageClusters within the same cluster. However, I don't think we're using it, and I don't believe it's relevant in this case because the StorageCluster is a child resource, not a parent. Therefore, we shouldn't need to delete the ownerReference.

Please correct me if I'm wrong.

subhamkrai:
> > We also need to scale down `odf-operator` when restoring storageCluster.
> 
> @subhamkrai The `StorageCluster` is an OCS CR, and I don't think the odf-operator is relevant here. As I understand it, the `odf-operator` is responsible for installing all operators, including OCS, Rook, NooBaa, Ceph CSI and etc..
> 
> There is also a `StorageSystem` CR managed by the `odf-operator`, which allows for deploying multiple StorageClusters within the same cluster. However, I don't think we're using it, and I don't believe it's relevant in this case because the StorageCluster is a child resource, not a parent. Therefore, we shouldn't need to delete the ownerReference.
> 
> Please correct me if I'm wrong.

Okay, I see scaling odf-operator is not required here.

travisn (approved):
Let's go ahead with this, given the testing looks good across several platforms and versions, at least for internal mode. Follow up is needed for testing with DR, etc. 
/lgtm

PR #137, opened 2025-04-01
Minor improvements in README.md

nirs:
- Remove the Requirements section since users know how to find the 
  required version using go.mod, and ODF cluster is not required to
  build the binary.
- Improve the heading to describe the content
- Simplify build instructions. `make build` works, but since it is the 
  default target (for good reason) we don't need to specify it. This is
  also the expected usage.
- Separate the build command to 2 lines for simplicity.
- Fix '/bin/' to 'bin/' adding missing space

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,14 +89,7 @@ These are the arguments that apply to all commands:
 
 ## Installation
 
-### Build from source
-
-#### Requirements
-
-- Go >= 1.22.0
-- ODF storage cluster should be installed.
-
-### Build and Run
+### Build and run from source
 
 1. Clone the repository
 
@@ -107,10 +100,11 @@ These are the arguments that apply to all commands:
 2. Change the directory and build the binary
 
     ```bash
-    cd odf-cli/ && make build
+    cd odf-cli/
+    make
     ```
 
-3. Use the binary present in the`/bin/` directory to run the commands
+3. Use the binary present in the `bin/` directory to run the commands
 
     ```bash
     ./bin/odf -h


subhamkrai (changes requested):
Since both are docs updates @nirs can you squash to a single commit? Thanks

nirs:
@subhamkrai thanks for reviewing. Squashed to one commit now.

subhamkrai (approved):

README.md
```diff
@@ -107,7 +107,8 @@ These are the arguments that apply to all commands:
 2. Change the directory and build the binary
 
     ```bash
-    cd odf-cli/ && make build
+    cd odf-cli/
+    make
```
subhamkrai, line 111:
```suggestion
    make build
```

subhamkrai (changes requested):

README.md
nirs: Why use `make build` when `make` wors?

I explained in the commit message why should use `make`:

```
- Simplify build instructions. `make build` works, but since it is the
  default target (for good reason) we don't need to specify it. This is
  also the expected usage.
```

subhamkrai: I agree it works but it is good practice to keep it complete so that new user don't get confused with only `make`

nirs: But most users are used to build projects like this, this how all projects work for many decades:

```console
make 
make install
make test
```

README.md
```diff
@@ -93,7 +93,7 @@ These are the arguments that apply to all commands:
 
 #### Requirements
```
travisn, line 94: I'd suggest we just remove this Requirements section. No need to maintain the go version, which easily gets outdated, and also it's not a requirement to have ODF installed just to build the tool.

README.md
travisn: Sounds good to me, `make` is commonly understood as the short version of `make build`

README.md
nirs: Right, this just creates more work for maintaining the readme and people know how to look up go version in go.mod.

@subhamkrai what do you think?

travisn (approved):
/lgtm

PR #138, opened 2025-04-04
build(deps): update ramenctl to v0.3.4

nirs:
Fixes includes in this version:
- https://github.com/RamenDR/ramenctl/issues/56
- https://github.com/RamenDR/ramenctl/pull/96
- https://github.com/RamenDR/ramenctl/pull/93

ramenctl requires now go 1.23.5, fixing CVE-2024-45336 (because of github.com/ramendr/ramen/e2e). `go mod tidy` updated the toolchain to go1.24.1, but I change it to go1.23.7 to minimize the chance of picking up packages that are not compatible with the downstream build.

With this version applications deployed by `odf dr test run` command can be viewed and manipulated in OpenShift console. This can be useful if the `odf dr test clean` was not run.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,12 @@
 module github.com/red-hat-storage/odf-cli
 
-go 1.23.0
+go 1.23.5
 
-toolchain go1.23.4
+toolchain go1.23.7
 
 require (
 	github.com/pkg/errors v0.9.1
-	github.com/ramendr/ramenctl v0.3.1
+	github.com/ramendr/ramenctl v0.3.4
 	github.com/red-hat-storage/ocs-operator/api/v4 v4.0.0-20240701091545-dfffbde82a9d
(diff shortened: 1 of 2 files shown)


subhamkrai (changes requested):
Changes looks good to me, can you just check once with build team that golang version are compatible with what we have?

nirs:
> Changes looks good to me, can you just check once with build team that golang version are compatible with what we have?

When we added the ramen requirement we check and we have go 1.23.6 in the downstream builders.

You can check this using:

```console
% podman run --rm --pull always --entrypoint /bin/go registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:rhel_9_golang_1.23 version
...
go version go1.23.6 (Red Hat 1.23.6-2.el9_6) linux/arm64
```

PR #141, opened 2025-04-07
docs: add download-odf-nightly.sh script

nirs:
The script downloads a nightly build and extracts the odf executable to
the current directory. This makes it easier to test the latest odf
build.

We use the v4.19 tag, pointing to the latest v4.19.z image.

Tested on:
- [x] macOS arm64 (using x86_64 build via Rosetta)
- [x] Linux x86_64

This nightly image does not contain other architectures. When we will have multiarch images the script should work on any machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,3 +109,15 @@ These are the arguments that apply to all commands:
     ```bash
     ./bin/odf -h
     ```
+
+### Download nightly build
+
+If you want to test the latest nightly build you can use the
+download-odf-nightly.sh script:
+
+```bash
+scripts/download-odf-nightly.sh v4.19
+```
+
+The command downloads the latest 4.19 build image and extracts the odf
+executable to the current directory.
--- a/scripts/download-odf-nightly.sh
+++ b/scripts/download-odf-nightly.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+# Download a nightly build of odf cli executable for the current platform and
+# architecture.
+
+set -e
+
+version="$1"
+
+if [ -z "$version" ]; then
+    cat <<EOF
+Usage: $0 VERSION
+
+Examples:
+
+  $0 v4.19
+
+EOF
+    exit 1
+fi
+
+image="quay.io/rhceph-dev/odf4-odf-cli-rhel9:$version"
+platform=$(uname)
+
+case $platform in
+Darwin)
+    platform=macosx
+    executable=odf
+    ;;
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -109,3 +109,15 @@ These are the arguments that apply to all commands:
...
     ```
+
+### Download nightly build
+
+If you want to test the latest nightly build you can use the
+download-odf-nightly.sh script:
+
+```bash
+./download-odf-nightly.sh 29
+```
```
subhamkrai, line 120:
```suggestion
```bash
./download-odf-nightly.sh 29
```
```

just wondering can we make this part of  new make command?

File: scripts/download-odf-nightly.sh [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+        ;;
+    esac
+    ;;
+*)
+    echo "unsupported platform $platform"
+    exit 1
+    ;;
+esac
+
+oc image extract quay.io/rhceph-dev/mcg-cli:$version-$build --file /usr/share/odf/$platform/$executable
```
nirs, line 45: I learned that we have this image for the latest release:
https://quay.io/repository/rhceph-dev/odf4-odf-cli-rhel9?tab=tags&tag=v4.19

Testing it now.

nirs:
@subhamkrai the script was simplified, no need to provide the build number now.

nirs:
@subhamkrai anything else needed? I think this ready.

scripts/download-odf-nightly.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+# Download a nightly build of odf cli executable for the current platform and
+# architecture.
+
+set -e
+
+image="quay.io/rhceph-dev/odf4-odf-cli-rhel9:v4.19"
```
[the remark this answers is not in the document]
nirs: Good idea! Previously we had a build number version which is hard to find, but the version is well known value, and this will enable using both v4.19 and v4.19.z as needed.

nirs:
Changes in lastest version:
- Require version (e.g. v4.19) so the user can download exact build, and we don't need to change the script for future releases (@travisn)
- Add Usage message
- Always download linux/amd64 image - latest images multiarch and we don't have images matching darwin/arm64. The linux/amd64 image includes odf executable for all platforms.

travisn (commented):
/lgtm

PR #161, opened 2025-06-02
command: add noobaa cli capability in the tool

subhamkrai:
this commit add the noobaa cli capability in the
cli tool to run all the noobaa command.

command structure
```
odf noobaa <args>
```


docs/noobaa.md
```diff
@@ -0,0 +1,45 @@
+# Noobaa Commands
+
+The Noobaa command supports all the noobaa cli arguments. See [noobaa cli](https://github.com/noobaa/noobaa-operator?tab=readme-ov-file#noobaa-cli)
```
travisn, line 3: Would the odf cli tool really support all the noobaa cli arguments? For example, do `install` and `upgrade` make sense in the cli tool? Shouldn't those be through standard ODF install/upgrade?

docs/noobaa.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+```bash
+Troubleshoot ODF noobaa
+
+Install:
+  install          Install the operator and create the noobaa system
+  upgrade          Upgrade the system, its components and CRDS
+  uninstall        Uninstall the operator and delete the system
+  status           Status of the operator and the system
```
travisn, line 18: Could we show `odf noobaa status` as the example? Or something else simple, so we don't have so much output that could become stale soon.

travisn (changes requested):

docs/noobaa.md
subhamkrai: Good question, I'm checking with the build team regarding the commands supports,.

docs/noobaa.md
subhamkrai: will do after testing

docs/noobaa.md
subhamkrai: Actually, there were few changes in upstream vs downstream commands in noobaa. See below output

docs/noobaa.md
```diff
@@ -0,0 +1,15 @@
+# Noobaa Commands
+
+The Noobaa command supports all the noobaa cli arguments available in odf.
+
+## Example
+
+```bash
+odf noobaa version
```
travisn, line 8: Could we run a noobaa command in the CLI to get basic validation that some command succeeds? Or maybe not if all commands require noobaa to be running.

subhamkrai: You mean to add in CI?

@travisn actually, we don't have any validation CI for downstream apart from test and build. We never agreed to run validatoin in github action with Kubernetes as we wanted to validate on OpenShift environment.

travisn (approved):
/lgtm

PR #171, opened 2025-08-04
DFBUGS-3525: Fix CVE x net 4.19

subhamkrai:
this commits update go/x/net and submariner to
their latest version i.e 0.42.0 and 0.21.0


subhamkrai:
Closing this bug as original issue is closed. Will re-open if issue comes later.

PR #177, opened 2025-08-20
Add new dr commands for for ODF 4.20

nirs:
This change adds new command off dr commands for ODF 4.20:

- test: Test disaster recovery flow in your clusters
- validate: Detect disaster recovery problems

Preview of new docs:
- https://github.com/nirs/odf-cli/blob/ramenctl-0.10.1/docs/dr.md#validate
- https://github.com/nirs/odf-cli/blob/ramenctl-0.10.1/docs/dr.md#gather


nirs:
@subhamkrai tested with #180.

PR #202, opened 2025-09-24
subvolume: hide subvolume delete command for now

subhamkrai:
subvolume delete command has some know issue, so till it is fixed let's hide that until we have a fix.


subhamkrai:
@travisn ^

PR #283, opened 2026-04-10
build: remove ob obc crds from gitignore

subhamkrai:
downstream build are not connected to network, and due to this build is failing to pull noobaa ob/obc crds at build time to fix this we need to have ob/obc present in the code base.


agarwal-mudit:
Merged to unblock the build, can revert back later if required.

PR #301, opened 2026-04-24
DFBUGS-6060: build: update grpc to latest 1.80.0

subhamkrai:
this commits udpate grpc to latest 1.80.0 and
it's requirements which also include go 1.24.0


go.mod
```diff
@@ -1,11 +1,8 @@
 module github.com/red-hat-storage/odf-cli
 
-go 1.23.1
-
-toolchain go1.23.7
+go 1.24.0
```
travisn, line 3: Does the 4.16 build support go 1.24?

subhamkrai: @travisn, I have asked the build team to update it to 1.24 and 1.25 as well, and I am waiting for QE before asking the build team to update.

subhamkrai: @travisn 1.24 should be available now in 4.16 confirmed with Boris on this

travisn (approved):
/lgtm

PR #302, opened 2026-04-24
build: update grpc to latest 1.80.0

subhamkrai:
this commits udpate grpc to latest 1.80.0 and
it's requirements which also include go 1.24.0.


subhamkrai:
Cosing the PR as no longer needed.

PR #303, opened 2026-04-27
object: add remote-obc with object enable/disable command

subhamkrai:
remote-obc as a arg is not printed as part of uses when `odf object enable -h -n <ns>` is ran see below output

```
odf object enable -h -n openshift-storage-client
Install the ObjectBucket and ObjectBucketClaim CRDs
Usage:
  odf object enable [flags]
Examples:
odf object enable remote-obc
```

it is only listed in the example section which might not be clear for users.

### DIFF
--- a/cmd/odf/object/object.go
+++ b/cmd/odf/object/object.go
@@ -20,7 +20,7 @@ var ObjectCmd = &cobra.Command{
 }
 
 var enableCmd = &cobra.Command{
-	Use:       "enable",
+	Use:       fmt.Sprintf("enable %s", remoteOBCArg),
 	Short:     "Install the ObjectBucket and ObjectBucketClaim CRDs",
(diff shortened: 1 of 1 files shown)


subhamkrai:
```yaml
./bin/odf object enable remote-obc -h -n something
Install the ObjectBucket and ObjectBucketClaim CRDs

Usage:
  odf object enable remote-obc [flags]

Examples:
odf object enable remote-obc

Flags:
  -h, --help   help for enable

Global Flags:
      --context string              Openshift context to use
      --kubeconfig string           Openshift config path
  -n, --namespace string            Openshift namespace where the StorageCluster CR is created (default "openshift-storage")
      --operator-namespace string   Openshift namespace where the ODF operator is running
```

travisn (approved):
/lgtm

PR #304, opened 2026-04-27
noobaa: use noobaa latest branch 5.22

subhamkrai:
let's use noobaa 5.22 branch which will be base for odf 4.22 but JFYI 5.22 is not release yet it just the branch name.


travisn (approved):
/lgtm

PR #321, opened 2026-05-12
build: bump go-jose/v3 to v3.0.5

subhamkrai:
Use go-jose/v3 v3.0.5 instead of v4.1.4 on this branch to avoid raising the module Go version (v4.1.4 requires go >= 1.24).


subhamkrai:
Cosing the PR as no longer needed.

PR #324, opened 2026-05-18
noobaa: update go mod to use latest 5.22 branch commit

subhamkrai:
there is no official release of noobaa 5.22, which will be base for 4.22 release. Let's use the latest 5.22 commit in the go mod.


travisn (commented):
@subhamkrai Is any change needed to the tag that we use for the noobaa crd install command? IIRC we are using the 5.21 tag?

subhamkrai:
> @subhamkrai Is any change needed to the tag that we use for the noobaa crd install command? IIRC we are using the 5.21 tag?

It is already there as https://github.com/red-hat-storage/odf-cli/blob/975bd95993ff0ca568efa55015c22c9bec3032c6/Makefile#L2 the branch was available when I doing the CRD changes

travisn (approved):
/lgtm

